PR #169, opened 2022-06-27
Fix MI-FNO link

AlexandreF-1qbit:
Several key points:
- We need to fetch the relevant fragment results with the `get_results` function in `qemist_client`. The `mo_coeffs` are not available anymore in the MI results.
- We don't need all the fragment results to compute a `FermionOperator`, but we do need all of them to redo the MI energy summation (epsilon are function of the correlation energies, and this info is only available in the fragment result dictionaries).
- I have chosen the `sto-3g` basis for the HF example in the notebook. It makes the QITE quick to run. A drawback of this is that there is no FNO truncation (all correction are 0.0).
- 3 files are important (the other ones are hdf5 and json files).


ValentinS4t1qbit (approved):
Good job Alexandre, thank you for fixing this

PR #174, opened 2022-06-30
Fix the need for spin with a qubit operator + pool of fermion operators

AlexandreF-1qbit:
`spin` was not defined with a `qubit_hamiltonian`, but it is needed to convert the pool of fermion operators to qubit operators.


ValentinS4t1qbit (approved):
Thank you fox fixing this, Alexandre

PR #177, opened 2022-07-21
Added restrictions for the frozen orbitals + FCISolver active space selection

AlexandreF-1qbit:
The main events:
- Restriction of the frozen orbitals (can't freeze orbitals with occupation == 1). There are inconsistencies with `pyscf` solvers when freezing half-filled MOs. At the moment, CASCI and MP2 do not work, CCSD does. Therefore, I think it is a good idea to raise an error until the issue https://github.com/pyscf/pyscf/issues/1331 is fixed/resolved.
- Changed qubits to spinorbitals in the fermion to qubit mapping function.
- Removed unused `frozen_orbitals` in `ADAPTSolver`.
- Fixed orbital selection in `FCISolver`.
- `VQESolver` raises an error when there is no var_params in the circuit.

### DIFF
--- a/tangelo/toolboxes/qubit_mappings/mapping_transform.py
+++ b/tangelo/toolboxes/qubit_mappings/mapping_transform.py
@@ -103,6 +103,9 @@ def fermion_to_qubit_mapping(fermion_operator, mapping, n_spinorbitals=None, n_e
     if mapping.upper() not in available_mappings:
         raise ValueError(f"Invalid mapping selection. Select from: {available_mappings}")
 
+    if mapping.upper in {"BK", "SCBK", "JKMN"} and n_spinorbitals is None:
+        raise ValueError(f"{mapping.upper()} requires n_spinorbitals to be set.")
+
     if up_then_down:
         if n_spinorbitals is None:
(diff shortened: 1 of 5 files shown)


tangelo/toolboxes/qubit_mappings/mapping_transform.py
```diff
@@ -112,11 +112,11 @@ def fermion_to_qubit_mapping(fermion_operator, mapping, n_spinorbitals=None, n_e
         qubit_operator = jordan_wigner(fermion_operator)
     elif mapping.upper() == "BK":
         if n_spinorbitals is None:
-            raise ValueError("Bravyi Kitaev requires specification of number of qubits.")
+            raise ValueError("Bravyi Kitaev requires specification of number of spin-orbitals.")
```
ValentinS4t1qbit, line 115: L115-119-130

lets just refactor them to have a single test before ths branches that says

```
if mapping.upper in {"BK", "SCBK", "JKMN"}:
   raise ValueError(f"{mapping.upper()} requires n_spinorbitals to be set"}
```

JamesB-1qbit (approved):

ValentinS4t1qbit (approved):

PR #178, opened 2022-07-22
lcu implementation

JamesB-1qbit:
This implements the Linear Combination of Unitaries algorithm. 
Both the standard form for any operator and the truncated taylor series form are included.

### DIFF
--- a/tangelo/toolboxes/circuits/lcu.py
+++ b/tangelo/toolboxes/circuits/lcu.py
@@ -67,7 +67,7 @@ def get_truncated_taylor_series(qu_op: QubitOperator, kmax: int, t: float, contr
     return amplified_lcu_circuit * rsteps
 
 
-def Uprepkl(qu_op_in: QubitOperator, kmax: int, t_in: float) -> Tuple[Circuit, List[QubitOperator], int]:
+def Uprepkl(qu_op: QubitOperator, kmax: int, t: float) -> Tuple[Circuit, List[QubitOperator], int]:
     """Generate Uprep circuit using qubit encoding defined in arXiv:1412.4687
     Args:
         qu_op (QubitOperator) :: The qubit operator to obtain the Uprep circuit for
@@ -80,13 +80,13 @@ def Uprepkl(qu_op_in: QubitOperator, kmax: int, t_in: float) -> Tuple[Circuit, L
     """
 
     # Incorporate sign of time into qubit operator.
-    qu_op = np.sign(t_in) * qu_op_in
-    t = abs(t_in)
+    qu_op_new = np.sign(t) * qu_op
+    t_new = abs(t)
 
     # remove the coefficient value and generate the list of QubitOperator with only its phase 1j, or -1j
     vector = list()
     unitaries = list()
-    for term, coeff in qu_op.terms.items():
+    for term, coeff in qu_op_new.terms.items():
         if np.abs(coeff.imag) > 1.e-7:
             raise ValueError(f"Only real qubit operators are allowed but term {term} has coefficient {coeff}")
         if coeff.real > 0:
@@ -106,8 +106,8 @@ def Uprepkl(qu_op_in: QubitOperator, kmax: int, t_in: float) -> Tuple[Circuit, L
     max_time_step = poly_roots[kmax] / vsum if kmax < 9 else np.log(2)
 
     # Calculate the number of time steps required and calculate the actual 1-norm for each time-step
-    time_steps = math.ceil(t / max_time_step)
-    vsum_with_t = sum(coeff * t / time_steps for coeff in vector)
+    time_steps = math.ceil(t_new / max_time_step)
+    vsum_with_t = sum(coeff * t_new / time_steps for coeff in vector)
     expvsum = sum((vsum_with_t) ** k / math.factorial(k) for k in range(0, kmax+1))
 
     # Generate vector v_i = sqrt(alpha_i)/np.sqrt(1-norm) with zeros padded to create a vector of length 2**n
@@ -122,7 +122,7 @@ def Uprepkl(qu_op_in: QubitOperator, kmax: int, t_in: float) -> Tuple[Circuit, L
     for k in range(kmax):
         qss.append(s.initializing_circuit())
         # Reindex to correct block of qubits for kth order of Taylor series
-        qss[-1].reindex_qubits(list(range(kmax + k * n_qubits, kmax + (k + 1) * n_qubits)))
+        qss[-1].reindex_qubits(list(range(kmax + k * n_qubits, kmax + (k+1) * n_qubits)))
         # Add control for the kth unary encoded qubits
         for gate in qss[-1]._gates:
             if gate.control is not None:
@@ -138,19 +138,47 @@ def Uprepkl(qu_op_in: QubitOperator, kmax: int, t_in: float) -> Tuple[Circuit, L
     # first qubit encodes extra identity term needed to ensure 1-norm = 2 such that
     # oblivious amplitude amplification is applicable
     kvec = np.zeros(2**(kmax + 1))
-    for k in range(0, kmax+1):
-        pos = int("0" + "0" * (kmax - k) + "1" * k, base=2)
-        kvec[pos] = (t / time_steps * vsum) ** k / math.factorial(k)
+    for k in range(kmax+1):
+        pos = int("0" + "0"*(kmax-k) + "1"*k, base=2)
+        kvec[pos] = (t_new / time_steps * vsum) ** k / math.factorial(k)
     kvec[0] += ((2 - expvsum) / 2)
     kvec[int("1" + "0" * kmax, base=2)] = ((2 - expvsum) / 2)
     kvec = np.sqrt(kvec) / np.sqrt(np.sum(kvec))
     ksv = StateVector(kvec, order="msq_first")
     kprep = ksv.initializing_circuit()
+    kprep = get_unary_prep(kvec, kmax)
 
     # shift encoding of extra identity term to last qubit
     kprep.reindex_qubits(list(range(kmax)) + [kmax + kmax * n_qubits])
 
     return kprep + qstot, unitaries, time_steps
+
+
+def get_unary_prep(kvec, kmax):
+    """Generate the prep circuit for the unary+ancilla part of the encoding. More efficient than using StateVector
+
+    Args:
+        kvec (array): Array representing the coefficients needed to generate
+        kmax (int): The Taylor series order
+
+    Returns:
+        Circuit : The unary encoding prep circuit
+    """
+
+    # Generate ancilla value and apply "X" so control is on other portion
+    val = kvec[int("1" + kmax*"0", base=2)]
+    gates = [Gate("RY", kmax, parameter=np.arcsin(val)*2), Gate("X", kmax)]
+
+    # Keep track of remaining value in constant c
+    c = np.cos(np.arcsin(val))
+    for i in range(0, kmax):
+        # Obtain new value to generate and rotate by np.arccos(val/c)
+        val = kvec[int("0" + "0"*(kmax-i) + "1"*i, base=2)]
+        gates += [Gate("CRY", i, control=[kmax]+list(range(i)), parameter=np.arccos(val/c)*2)]
+        c *= np.sin(np.arccos(val/c))
+    gates += [Gate("X", kmax)]
+
+    return Circuit(gates)
 
 
 def USelectkl(unitaries: List[QubitOperator], n_qubits_sv: int, kmax: int, control: Union[int, List[int]] = None) -> Circuit:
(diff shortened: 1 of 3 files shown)


AlexandreF-1qbit (changes requested):
I am good with the most of it. Here are some general questions:
- Could `get_truncated_taylor_series` be in a different file as it seems different vs the `LCU`?
- Is there any justification for the use of `vec = (vecs[:, 0] + vecs[:, 1])/np.sqrt(2)` as the initial state in the tests?

tangelo/toolboxes/circuits/lcu.py
```diff
@@ -0,0 +1,308 @@
...
+    expvsum = sum((vsum_with_t) ** k / math.factorial(k) for k in range(0, kmax+1))
+
+    # Generate vector v_i = sqrt(alpha_i)/np.sqrt(1-norm) with zeros padded to create a vector of length 2**n
+    vector = np.sqrt(vector) / np.sqrt(vsum)
+    n_qubits = math.ceil(math.log2(num_terms))
+    newvec = np.zeros(2 ** n_qubits)
+    newvec[:num_terms] = vector
+
+    # Calculate circuit that generates vector and apply controls for taylor series
+    s = StateVector(newvec, order="msq_first")
```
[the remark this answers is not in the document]
JamesB-1qbit: I hadn't thought about this in a while, but this was good to test. The "msq_first" relates to how I label the USelect operation so it is fine for any backend. (i.e. The "011" vs "110" control labels.)

But thanks to this comment, the testing is now improved such that both "cirq" and "qulacs" are tested which have different ordering. "qiskit" does not support general multiqubit gates so is not currently supported. I have now realized that "qiskit" does have a multi controlled X gate (https://qiskit.org/documentation/apidoc/circuit_library.html) which we could use to support all multi-controlled gates in theory if we care.

tangelo/toolboxes/circuits/lcu.py
```diff
@@ -0,0 +1,310 @@
...
+    newvec = np.zeros(2 ** n_qubits)
+    newvec[:num_terms] = vector
+
+    # Calculate circuit that generates vector and apply controls for Taylor series
+    s = StateVector(newvec, order="msq_first")
+    qss = list()
+    for k in range(kmax):
+        qss.append(s.initializing_circuit())
+        # Reindex to correct block of qubits for kth order of Taylor series
+        qss[-1].reindex_qubits(list(range(kmax + k * n_qubits, kmax + (k + 1) * n_qubits)))
```
ValentinS4t1qbit, line 125: I recommend `(k+1)` without spaces for readability.

tangelo/toolboxes/circuits/lcu.py
```diff
@@ -0,0 +1,310 @@
...
+                gate.control = [k]
+    qstot = Circuit()
+    for qs in qss:
+        qstot += qs
+
+    # Calculate coefficients for unary encoding k="0"+"0"*(kmax-k)*"1"*k where
+    # first qubit encodes extra identity term needed to ensure 1-norm = 2 such that
+    # oblivious amplitude amplification is applicable
+    kvec = np.zeros(2**(kmax + 1))
+    for k in range(0, kmax+1):
```
ValentinS4t1qbit, line 141: `range(kmax+1)`

tangelo/toolboxes/circuits/lcu.py
```diff
@@ -0,0 +1,310 @@
...
+        pos = int("0" + "0" * (kmax - k) + "1" * k, base=2)
```
ValentinS4t1qbit, line 142: For readability, its useful to vary the number of spaces. Spacing everything equally (whether no space of 1 space) makes it almost equally hard to figure out operation priority. Here's my personal preference here, for example; it helps me see blocks more easily.

`pos = int("0" + "0"*(kmax-k) + "1"*k, base=2)`

tangelo/toolboxes/circuits/lcu.py
```diff
@@ -0,0 +1,310 @@
...
+    Args:
+        unitaries (list[QubitOperator]): The list of unitaries that defines the U_{Select} operation
+        n_qubits_sv (int): The number of qubits in the statevector register
+        kmax (int): The maximum Taylor series order
+        control (int or list[int]): Control qubits for operation
+    Returns:
+        Circuit: The circuit that implements the truncated Taylor series U_{Select} operation
+    """
+
+    n_qubits_u = math.ceil(math.log2(len(unitaries)))
```
ValentinS4t1qbit, line 171: do we need a check on the resulting value, can `len(unitaries)` be anything or should it be a power of 2?

tangelo/toolboxes/circuits/lcu.py
```diff
@@ -0,0 +1,310 @@
...
+        Circuit: The circuit that generates the sign flip on |0>^n
+    """
+    if control is not None:
+        fcontrol_list = control if isinstance(control, list) else [control]
+    else:
+        fcontrol_list = []
+    gate_list = []
+
+    x_ladder = [Gate("X", q) for q in qubit_list]
+    fcontrol_list += [q for q in qubit_list[:-1]]
```
ValentinS4t1qbit, line 226: Can't we just do `fcontrol_list = qubit_list[:-1]` ?
This is a slicing operation so I would expect this to create a copy of all the element of that list except the last one.

You're only using this at L229 and you're not modifying it, so maybe we don't need 226 we can just go to L229 and write `control=qubit_list[:-1]`

tangelo/toolboxes/circuits/lcu.py
```diff
@@ -0,0 +1,310 @@
...
+    vector = list()
+    max_qu_op = count_qubits(qu_op)
+    for term, coeff in qu_op.terms.items():
+        acoeff = np.abs(coeff)
+        vector += [acoeff]
+        unitaries += [QubitOperator(term, coeff / acoeff)]
+    vsum = sum(vector)
+
+    # Check that 1-norm is less than 2 and add and subtract terms proportional to the identity to obtain 1-norm equal to 2
+    if vsum > 2:
```
ValentinS4t1qbit, line 259: I think you may want to account for a machine precision epsilon here (e.g what if the sum is 2.00000000001?)

`if vsum > 2 + 1e-k:` (I let you decide what is a reasonable value for `k`)

tangelo/toolboxes/circuits/tests/test_lcu.py
```diff
@@ -0,0 +1,187 @@
...
+        time = 1.9
+        exact = expm(-1j*ham*time)@vec
+
+        for backend in backends:
+            sim = Simulator(backend)
+            statevector_order = backend_info[backend]["statevector_order"]
+            sv = StateVector(vec, order=statevector_order)
+            sv_circuit = sv.initializing_circuit()
+
+            # Tested for up to k = 5 but 5 is very slow.
```
ValentinS4t1qbit, line 60: It would be good to know why it's slow (e.g what part of the computation scales "badly" with k).
You could very quickly print 2-3 timers that use functions you think are responsible to get that info.

Maybe the answer is simply that simulators are just spending a lot of time simulating circuits.

tangelo/toolboxes/circuits/tests/test_lcu.py
```diff
@@ -0,0 +1,187 @@
...
+            sv = StateVector(vec, order=statevector_order)
+            sv_circuit = sv.initializing_circuit()
+
+            taylor_circuit = get_lcu_circuit(exp_qu_op)
+            _, v = sim.simulate(sv_circuit + taylor_circuit, return_statevector=True)
+            v = v.reshape([4, len_ancilla])[:, 0] if statevector_order == "lsq_first" else v.reshape([len_ancilla, 4])[0, :]
+            self.assertAlmostEqual(1, np.abs(v.conj().dot(exact)), delta=1.e-3)
+
+        # Test return of ValueError if 1-norm is greater than 2.
+        self.assertRaises(ValueError, get_lcu_circuit, exp_qu_op+5)
```
ValentinS4t1qbit, line 101: Is this consistent with what L100 announced ?

tangelo/toolboxes/circuits/tests/test_lcu.py
```diff
@@ -0,0 +1,187 @@
...
+        """
+
+        # Generate qubit operator with state 9 having eigenvalue 0.25
+        qu_op = (QubitOperator("X0 X1", 0.125) + QubitOperator("Y1 Y2", 0.125) + QubitOperator("Z2 Z3", 0.125)
+                 + QubitOperator("", 0.125))
+
+        ham_mat = get_sparse_operator(qu_op).toarray()
+        _, wavefunction = np.linalg.eigh(ham_mat)
+
+        qu_op = qu_op
```
ValentinS4t1qbit, line 115: ?

tangelo/toolboxes/circuits/tests/test_lcu.py
```diff
@@ -0,0 +1,187 @@
...
+
+        freqs, _ = sim_cirq.simulate(pe_circuit, initial_statevector=wave_9)
+
+        # Trace out all but final 3 indices
+        trace_freq = dict()
+        for key, value in freqs.items():
+            trace_freq[key[-3:]] = trace_freq.get(key[-3:], 0) + value
+
+        # State 9 has eigenvalue 0.25 so return should be 010 (0*1/2 + 1*1/4 + 0*1/8)
+        self.assertAlmostEqual(trace_freq["010"], 1.0, delta=0.0001)
```
ValentinS4t1qbit, line 136: `delta=1e-4`

tangelo/toolboxes/circuits/tests/test_lcu.py
```diff
@@ -0,0 +1,187 @@
...
+        """
+
+        # Generate qubit operator with state 9 having eigenvalue 0.25
+        qu_op = (QubitOperator("X0 X1", 0.125) + QubitOperator("Y1 Y2", 0.125) + QubitOperator("Z2 Z3", 0.125)
+                 + QubitOperator("", 0.125))
+
+        ham_mat = get_sparse_operator(qu_op).toarray()
+        _, wavefunction = np.linalg.eigh(ham_mat)
+
+        qu_op = qu_op
```
ValentinS4t1qbit, line 150: ?

tangelo/toolboxes/circuits/tests/test_lcu.py
```diff
@@ -0,0 +1,187 @@
...
+        for i in range(n_extra_qubits+3):
+            wave_9 = np.kron(wave_9, np.array([1, 0]))
+
+        # Phase estimation register in reversed order
+        qubit_list = list(reversed(range(4 + n_extra_qubits, 4 + n_extra_qubits + 3)))
+
+        # Build phase estimation circuit and simulate
+        pe_circuit = get_qft_circuit(qubit_list)
+        for i, qubit in enumerate(qubit_list):
+            pe_circuit += get_lcu_circuit(exp_qu_op, control=qubit) * 2 ** i * 6
```
ValentinS4t1qbit, line 172: I suggest `6 * 2**i * get_lcu_circuit(...)`, which is typically how I'd write formulas on paper.
This assumes btw the `*` operator works both ways, which I think it does?

tangelo/toolboxes/circuits/tests/test_lcu.py
```diff
@@ -0,0 +1,187 @@
...
+
+        freqs, _ = sim_cirq.simulate(pe_circuit, initial_statevector=wave_9)
+
+        # Trace out all but final 3 indices
+        trace_freq = dict()
+        for key, value in freqs.items():
+            trace_freq[key[-3:]] = trace_freq.get(key[-3:], 0) + value
+
+        # State 9 has eigenvalue 0.25 so return should be 010 (0*1/2 + 1*1/4 + 0*1/8)
+        self.assertAlmostEqual(trace_freq["010"], 1.0, delta=0.0001)
```
ValentinS4t1qbit, line 183: `delta=1e-4`

ValentinS4t1qbit (changes requested):
A few things. Good work, you make solid use of the tangelo objects and numpy, your documentation and flow looks great as well.

tangelo/toolboxes/circuits/tests/test_lcu.py
JamesB-1qbit: The main reason that the algorithm slows done for each added k is that the number of qubits increases by 4 for each additional k. So k=5 has 23 qubits compared to 19 qubits for k=4. 

The scaling of the gates was also exponential but your question made me realize that I had an inefficiency in the code. I now added a `get_unary_prep` function that does the unary encoding in linear gates so the test with `k=5` takes 30s using qulacs. I don't think n=5 is necessary for testing purposes.

tangelo/toolboxes/circuits/lcu.py
JamesB-1qbit: The `math.ceil` rounds up to the nearest qubit number greater than `math.log2(len(unitaries))` so the function takes care of the power of 2.

JamesB-1qbit:
> I am good with the most of it. Here are some general questions:
> 
> * Could `get_truncated_taylor_series` be in a different file as it seems different vs the `LCU`?
> * Is there any justification for the use of `vec = (vecs[:, 0] + vecs[:, 1])/np.sqrt(2)` as the initial state in the tests?

* They are both LCU based algorithms. They just use a different encoding for the ancilla qubits.
* No, just something that is not an eigenvector of the Hamiltonian.

tangelo/toolboxes/circuits/tests/test_lcu.py
ValentinS4t1qbit: awesome, glad the comment was useful somehow.

ValentinS4t1qbit (approved):
Good stuff !

AlexandreF-1qbit (approved):

Issue #149, opened 2022-04-19
[BUG] Add a check for number of qubits involved in supported gates


elloyd-1qbit:
Small issue. An error should be raised when the number of target qubits does not match how many are involved in the gate. It looks like this check only exists for gates with a control qubit.
```
from tangelo.linq import Circuit, Gate

circ=Circuit()

swap_1=Gate('SWAP',[0])
swap_3=Gate('SWAP',[0,1,2])
circ.add_gate(swap_1)
circ.add_gate(swap_3)

XX_1=Gate('XX', [0])
XX_3=Gate('XX', [0,1,3])

circ.add_gate(XX_1)
circ.add_gate(XX_3)

RX_3=Gate('RX', [0,1,3])
circ.add_gate(RX_3)

print(circ)

```
returns:

```
SWAP      target : [0]   
SWAP      target : [0, 1, 2]   
XX        target : [0]   
XX        target : [0, 1, 3]   
RX        target : [0, 1, 3] 
```

AlexandreF-1qbit:
Close with #180.

PR #180, opened 2022-07-25
Check for the number of target qubits

AlexandreF-1qbit:
Naive attempt at fixing https://github.com/goodchemistryco/Tangelo/issues/149. It checks the number of target qubits for some gates. It does not prevent multicontrolled-qubit gates nor custom gates.

### DIFF
--- a/tangelo/linq/gate.py
+++ b/tangelo/linq/gate.py
@@ -99,7 +99,7 @@ class Gate(dict):
         else:
             n_targets = len(target)
         if len(target) != n_targets:
-            raise ValueError(f"Gate {g.name}: expected {n_targets} target qubits, but got {target}")
+            raise ValueError(f"Gate {name}: expected {n_targets} target qubits, but got {target}")
 
         self.__dict__ = {"name": name, "target": target, "control": control,
                          "parameter": parameter, "is_variational": is_variational}
--- a/tangelo/linq/gate.py
+++ b/tangelo/linq/gate.py
@@ -25,9 +25,9 @@ ONE_QUBIT_GATES = {"H", "X", "Y", "Z", "S", "T", "RX", "RY", "RZ", "PHASE"}
 TWO_QUBIT_GATES = {"CNOT", "CX", "CY", "CZ", "CRX", "CRY", "CRZ", "CPHASE", "XX", "SWAP"}
 THREE_QUBIT_GATES = {"CSWAP"}
 
-ONE_QUBIT_TARGET_GATES = {"H", "X", "Y", "Z", "S", "T", "RX", "RY", "RZ", "PHASE",
+ONE_TARGET_GATES = {"H", "X", "Y", "Z", "S", "T", "RX", "RY", "RZ", "PHASE",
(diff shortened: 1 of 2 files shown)


JamesB-1qbit (approved):

tangelo/linq/gate.py
```diff
@@ -84,6 +89,18 @@ def check_qubit_indices(qubit_list, label):
...
+        # the X_QUBIT_GATES sets), ignore the check. We also do not prevent
+        # the creation of multi-controlled gates (begins with "C").
+        if name.upper() in ONE_QUBIT_TARGET_GATES:
+            n_qubits = 1
+        elif name.upper() in TWO_QUBIT_TARGET_GATES:
+            n_qubits = 2
+        else:
+            n_qubits = len(target)
+        if len(target) != n_qubits:
+            raise ValueError(f"There are too many target/control qubits for this gate.")
```
ValentinS4t1qbit, line 102: 1. Replace "this gate" by the actual name of the gate, let's be useful there. I'd rather use `n_targets` than `n_qubits` as we use `n_qubits` many other places. We could say something like.

`f"Gate {g.name} : expected {n_qubits} target qubits, but got {target}"`

2. Your current code only seems to check that the number of targets is good, I'm not sure why your current error message says 'control' as well. Should there be a check for control, or none at all ?

tangelo/linq/gate.py
```diff
@@ -24,6 +24,11 @@
 ONE_QUBIT_GATES = {"H", "X", "Y", "Z", "S", "T", "RX", "RY", "RZ", "PHASE"}
 TWO_QUBIT_GATES = {"CNOT", "CX", "CY", "CZ", "CRX", "CRY", "CRZ", "CPHASE", "XX", "SWAP"}
 THREE_QUBIT_GATES = {"CSWAP"}
+
+ONE_QUBIT_TARGET_GATES = {"H", "X", "Y", "Z", "S", "T", "RX", "RY", "RZ", "PHASE",
```
ValentinS4t1qbit, line 28: I suggest `ONE_TARGET_GATES` and `TWO_TARGET_GATES`.

tangelo/linq/tests/test_gates.py
```diff
@@ -94,6 +94,15 @@ def test_gate_equality_modulo_twopi(self):
         self.assertTrue(g1 == g3)
         self.assertTrue(g2 == g3)
 
+    def test_too_many_qubits_on_gates(self):
+        """ Test the behavior when too many qubits are selected for a gate. """
+
+        # Trying to create a Hadamard gate acting on qubits 0 and 1.
```
ValentinS4t1qbit, line 100: `Try`. Do not conjugate verbs. Same with line 103

ValentinS4t1qbit (changes requested):
Thank you for addressing this long-lasting bug, which gives us the opportunity to show on Github that we follow up on those !

tangelo/linq/gate.py
AlexandreF-1qbit: Oh this is a mistake as I changed my mind last minute on the control check.

ValentinS4t1qbit (approved):

ValentinS4t1qbit (approved):

PR #181, opened 2022-07-27
Improve usability of MIFNOHelper

AlexandreF-1qbit:
Here is the summary:
- Users can now select a fragment nested dictionary with the [] operator.
- Definition of a method to output the number of electrons ans spinorbitals in a fragment (a must for fermion to qubit mapping).
- Added the use of `n_electrons_spinorbs` in the example `mifno` notebook.


JamesB-1qbit (approved):
Looks good

ValentinS4t1qbit (approved):

ValentinS4t1qbit (approved):

PR #183, opened 2022-07-29
qiskit noisy device

JamesB-1qbit:
Previous branch that was stale. An example of how to use is in examples/linq/qiskit-noisy-device.ipynb.

### DIFF
--- a/tangelo/linq/simulator.py
+++ b/tangelo/linq/simulator.py
@@ -270,7 +270,7 @@ class Simulator:
             backend = AerSimulator.from_backend(device_backend)
 
             if meas_mitt:
-                meas_calibs, state_labels= complete_meas_cal(qr=q, circlabel='mcal')
+                meas_calibs, state_labels = complete_meas_cal(qr=q, circlabel='mcal')
                 t_qc = qiskit.transpile(meas_calibs, backend, initial_layout=virtual_to_physical)
                 qobj = qiskit.assemble(t_qc, shots=10000)
                 cal_results = backend.run(qobj, shots=10000).result()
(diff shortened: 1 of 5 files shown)


tangelo/linq/simulator.py
```diff
@@ -86,7 +87,7 @@ def __init__(self, target=default_simulator, n_shots=None, noise_model=None):
         if not self.n_shots and (not self.statevector_available or self._noise_model):
             raise ValueError("A number of shots needs to be specified.")
 
-    def simulate(self, source_circuit, return_statevector=False, initial_statevector=None):
+    def simulate(self, source_circuit, return_statevector=False, initial_statevector=None, qubits_to_use=None, opt_level=1, meas_mitt=False):
```
ValentinS4t1qbit, line 90: I don't think these options belong here at first glance, I think they should be attributes of a backend object.
We want ideally a clear separation between the object that define what the computation is, and whatever is backend related.

tangelo/linq/simulator.py
```diff
@@ -230,6 +235,80 @@ def simulate(self, source_circuit, return_statevector=False, initial_statevector
                            if v > self.freq_threshold}
             return (frequencies, None)
 
+        elif self._target == "qiskit_device":
+            import qiskit
+            from qiskit.test import mock
+            from qiskit.providers.aer import AerSimulator
+            from qiskit.ignis.mitigation.measurement import complete_meas_cal, CompleteMeasFitter
```
ValentinS4t1qbit, line 242: Qiskit Ignis has been deprecated in favour of qiskit_experiments etc (see readme at https://github.com/Qiskit/qiskit-ignis)

tangelo/linq/simulator.py
```diff
@@ -264,7 +343,7 @@ def simulate(self, source_circuit, return_statevector=False, initial_statevector
 
             return (frequencies, np.array(self._current_state)) if return_statevector else (frequencies, None)
 
-    def get_expectation_value(self, qubit_operator, state_prep_circuit, initial_statevector=None):
+    def get_expectation_value(self, qubit_operator, state_prep_circuit, initial_statevector=None, qubits_to_use=None, opt_level=1, meas_mitt=False):
```
ValentinS4t1qbit, line 346: Same comment than for `simulate` regarding the arguments here

PR #193, opened 2022-08-15
copy + get_resources + invertible swap gate fix

JamesB-1qbit:
A bunch of small fixes.

1. copy function in circuit
2. get_resources update for VQESolver and SA_VQESolver
3. Missing comma in list of invertible gates meant swap gate could not be inverted.

### DIFF
--- a/tangelo/algorithms/variational/sa_vqe_solver.py
+++ b/tangelo/algorithms/variational/sa_vqe_solver.py
@@ -220,3 +220,23 @@ class SA_VQESolver(VQESolver):
             print(f"\tEnergy = {energy:.7f} ")
 
         return energy
+
+    def get_resources(self):
+        """Estimate the resources required by SA-VQE, with the current ansatz. This
+        assumes "build" has been run, as it requires the ansatz circuit and the
+        qubit Hamiltonian. Return information that pertains to the user, for the
+        purpose of running an experiment on a classical simulator or a quantum
+        device.
+        """
+
+        resources = dict()
+        resources["qubit_hamiltonian_terms"] = (len(self.qubit_hamiltonian.terms) + len(self.deflation_circuits))*self.n_states
+        circuit = (self.reference_circuits[0] + self.ansatz.circuit + self.deflation_circuits[0] if self.deflation_circuits else
+                   self.reference_circuits[0] + self.ansatz.circuit)
+        resources["circuit_width"] = circuit.width
+        resources["circuit_gates"] = circuit.size
+        # For now, only CNOTs supported.
+        resources["circuit_2qubit_gates"] = circuit.counts.get("CNOT", 0)
+        resources["circuit_var_gates"] = len(self.ansatz.circuit._variational_gates)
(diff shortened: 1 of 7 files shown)


tangelo/linq/tests/test_gates.py
```diff
@@ -52,6 +52,11 @@ def test_some_gates_inverse(self):
         H_gate_inverse = Gate("H", 2)
         self.assertEqual(H_gate.inverse().__str__(), H_gate_inverse.__str__())
 
+        # Create a SWAP gate acting on qubits 1 and 2
+        H_gate = Gate("SWAP", [1, 2])
+        H_gate_inverse = Gate("SWAP", [1, 2])
+        self.assertEqual(H_gate.inverse().__str__(), H_gate_inverse.__str__())
```
ValentinS4t1qbit, line 58: I don't know how old these tests are, but we should use Gate equality operators (==) to compare gates. Would you please quickly rewrite those four?
(also calling your SWAP gate "H_gate" is not the best.)

I think they could look something like that:

```
H_gate = Gate(...)
self.assertEqual(H_gate.inverse(), H_gate)

RZ_gate = Gate(...)
self.assertEqual(RZ_gate.inverse(), Gate('RZ', .... <change parameter>))
```

tangelo/linq/circuit.py
```diff
@@ -128,6 +128,10 @@ def is_mixed_state(self):
         """
         return "MEASURE" in self.counts
 
+    def copy(self):
+        """Return a copy of circuit"""
```
ValentinS4t1qbit, line 132: Let's be clear here with the user: this is a deep copy.

tangelo/algorithms/variational/sa_vqe_solver.py
```diff
@@ -220,3 +220,23 @@ def energy_estimation(self, var_params):
...
+        """
+
+        resources = dict()
+        resources["qubit_hamiltonian_terms"] = (len(self.qubit_hamiltonian.terms) + len(self.deflation_circuits))*self.n_states
+        circuit = (self.reference_circuits[0] + self.ansatz.circuit + self.deflation_circuits[0] if self.deflation_circuits else
+                   self.reference_circuits[0] + self.ansatz.circuit)
+        resources["circuit_width"] = circuit.width
+        resources["circuit_gates"] = circuit.size
+        # For now, only CNOTs supported.
+        resources["circuit_2qubit_gates"] = circuit.counts.get("CNOT", 0)
```
ValentinS4t1qbit, line 239: We'll have to think about this one later. Now that we have defined sets of gates in https://github.com/goodchemistryco/Tangelo/blob/main/tangelo/linq/gate.py , we may have to sum over more gates rather than CNOTs.

ValentinS4t1qbit (changes requested):
A bit early as this is still a draft but here's some early feedback

ValentinS4t1qbit (approved):
Thank you James !

ValentinS4t1qbit (approved):
Thank you James, sorry for making a change hastily to your work: that was a mistake.

PR #194, opened 2022-08-15
Iterable circuit

AlexandreF-1qbit:
Allow users to traverse a circuit object, in order to iterate over gates.

```python
# Create circuit with a dummy list of gates
gates = ['H', 'X', 'CNOT', 'RZ']
c = Circuit(gates)

# Now, traversing the circuit is equivalent to iterating over gates
for g in c:
    print(g)
```

### DIFF
--- a/tangelo/linq/circuit.py
+++ b/tangelo/linq/circuit.py
@@ -102,6 +102,10 @@ class Circuit:
         """
         return iter(self._gates)
 
+    def __next__(self):
+        """Define the next function when calling next(Circuit). """
+        return next(self._gates)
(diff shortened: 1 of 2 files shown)


ValentinS4t1qbit (changes requested):
Thank you: let's add a test.

Maybe simply something that asserts equality between the two following lists:
`[g for g in c]` and `c._gates`

JamesB-1qbit (approved):

ValentinS4t1qbit (approved):
Thanks Alexandre !

PR #197, opened 2022-08-16
Excited states

JamesB-1qbit:
Test of excited_states.ipynb testing after review fixes.


ValentinS4t1qbit (approved):
great work James, i think this is a very interesting notebook

PR #203, opened 2022-08-26
translate qiskit.QuantumCircuit to Circuit 

JamesB-1qbit:
Change `translate_qiskit` such that a qiskit.QuantumCircuit with supported gates can be translated to Circuit format.

### DIFF
--- a/tangelo/linq/gate.py
+++ b/tangelo/linq/gate.py
@@ -125,7 +125,10 @@ class Gate(dict):
 
         ds, do = self.__dict__, other.__dict__
 
-        if any(ds[k] != do[k] for k in ds if k != "parameter"):
+        # Check CNOT CX Separately
+        ignore_list = ["name", "parameter"] if ds["name"] in ["CNOT", "CX"] and do["name"] in ["CNOT", "CX"] else ["parameter"]
+
+        if any(ds[k] != do[k] for k in ds if k not in ignore_list):
             return False
 
         parameter = round(ds["parameter"] % (2 * pi), 7) if isinstance(ds["parameter"], (float, int)) else ds["parameter"]
--- a/tangelo/linq/gate.py
+++ b/tangelo/linq/gate.py
@@ -125,7 +125,8 @@ class Gate(dict):
 
         ds, do = self.__dict__, other.__dict__
 
-        # Check CNOT CX Separately
+        # CNOT and CX gates are equivalent. If both gates are either CNOT or CX, the names are equivalent and can
+        # be ignored for equivalence checks later.
         ignore_list = ["name", "parameter"] if ds["name"] in ["CNOT", "CX"] and do["name"] in ["CNOT", "CX"] else ["parameter"]
 
         if any(ds[k] != do[k] for k in ds if k not in ignore_list):
--- a/tangelo/linq/translator/__init__.py
+++ b/tangelo/linq/translator/__init__.py
@@ -15,7 +15,7 @@
 from tangelo.helpers.utils import installed_backends
 
 from .translate_braket import translate_braket, get_braket_gates
-from .translate_qiskit import translate_qiskit, get_qiskit_gates
+from .translate_qiskit import translate_qiskit, get_qiskit_gates, translate_c_from_qiskit, translate_c_to_qiskit
 from .translate_qulacs import translate_qulacs, get_qulacs_gates
 from .translate_cirq import translate_cirq, get_cirq_gates
 from .translate_json_ionq import translate_json_ionq, get_ionq_gates
--- a/tangelo/linq/translator/translate_qiskit.py
+++ b/tangelo/linq/translator/translate_qiskit.py
@@ -25,7 +25,6 @@ The module also enables bidirectional conversion between qiskit and Tangelo
 qubit operators (linear combination of Pauli operators)
 """
 from tangelo.linq import Circuit, Gate
-
 from tangelo.toolboxes.operators import QubitOperator
 from tangelo.linq.helpers import pauli_of_to_string, pauli_string_to_of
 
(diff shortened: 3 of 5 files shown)


tangelo/linq/gate.py
```diff
@@ -125,7 +125,10 @@ def __eq__(self, other):
 
         ds, do = self.__dict__, other.__dict__
 
-        if any(ds[k] != do[k] for k in ds if k != "parameter"):
+        # Check CNOT CX Separately
```
ValentinS4t1qbit, line 128: Can you be a bit more descriptive with that comment ?
What this code means to me is "Consider CX and CNOT gates to be the same"

tangelo/linq/translator/translate_qiskit.py
```diff
@@ -24,6 +24,7 @@
 The module also enables bidirectional conversion between qiskit and Tangelo
 qubit operators (linear combination of Pauli operators)
 """
+from tangelo.linq import Circuit, Gate
```
ValentinS4t1qbit, line 27: I'm not sure why L28 is skipped here

ValentinS4t1qbit (changes requested):
It looks like the functionality is there, thank you ! I think we need to change a few things however.

Can I please ask you to add a "performance test" (similar to what we did in https://github.com/goodchemistryco/Tangelo/blob/main/tangelo/linq/tests/test_translator_qubitop.py) to check how much time each direction take for a decently large quantum circuit. You can generate an arbitrary quantum circuit with a relevant size (100K gates?) and maybe time both ways.

As long as the test takes less than 30 secs it can be included inautomated tests, otherwise we can `unittest.skip` it

tangelo/linq/translator/translate_qiskit.py
```diff
@@ -103,6 +115,48 @@ def translate_qiskit(source_circuit):
...
             raise ValueError(f"Gate '{gate.name}' not supported on backend qiskit")
+
+    return target_circuit
+
+
+def translate_c_from_qiskit(source_circuit) -> Circuit:
+    """Take in a qiskit.QuantumCircuit, return an equivalent Circuit
+
+    Args:
+        source_circuit (qiskit.QuantumCircuit): quantum circuit in the abstract qiskit.QuantumCircuit format.
```
ValentinS4t1qbit, line 126: "abstract" is what i used to qualify our (Tangelo's) circuits, so I dont think this sentence should include it

tangelo/linq/translator/translate_qiskit.py
```diff
@@ -103,6 +115,48 @@ def translate_qiskit(source_circuit):
             (GATE_QISKIT[gate.name])(target_circuit, gate.target[0], gate.target[0])
         else:
             raise ValueError(f"Gate '{gate.name}' not supported on backend qiskit")
+
+    return target_circuit
+
+
+def translate_c_from_qiskit(source_circuit) -> Circuit:
+    """Take in a qiskit.QuantumCircuit, return an equivalent Circuit
```
ValentinS4t1qbit, line 123: Circuit -> Tangelo Circuit object (to lift any ambiguity)

tangelo/linq/translator/translate_qiskit.py
```diff
@@ -103,6 +115,48 @@ def translate_qiskit(source_circuit):
...
+
+    Args:
+        source_circuit (qiskit.QuantumCircuit): quantum circuit in the abstract qiskit.QuantumCircuit format.
+
+    Returns:
+        Circuit: the corresponding quantum Circuit.
```
ValentinS4t1qbit, line 129: in Tangelo format

tangelo/linq/translator/translate_qiskit.py
```diff
@@ -103,6 +115,48 @@ def translate_qiskit(source_circuit):
...
+        elif name in {"SWAP"}:
+            gates += [Gate(name, [qi[gate.qubits[0]].index, qi[gate.qubits[1]].index])]
+        elif name in {"CSWAP"}:
+            gates += [Gate(name, [qi[gate.qubits[1]].index, qi[gate.qubits[2]].index], control=qi[gate.qubits[0]].index)]
+        elif name in {"XX"}:
+            gates += [Gate(name, [qi[gate.qubits[0]].index, qi[gate.qubits[1]].index], parameter=gate.operation.params[0])]
+        elif name in {"MEASURE"}:
+            gates += [Gate(name, qi[gate.qubits[0]].index)]
+        else:
+            raise ValueError(f"Gate '{gate.name}' not supported on backend qiskit")
```
ValentinS4t1qbit, line 157: Shouldn't this say that the gate.name is not support in Tangelo instead ? Since gate is froma  source circuit in qiskit format, qiskit should obviously support it.

Also, any name is supported in our Tangelo Circuit. We don't check for names since it is an"abstract" format (POTATO gate example etc). I'm not sure this check is warranted at all

tangelo/linq/translator/translate_qiskit.py
```diff
@@ -65,27 +65,39 @@ def get_qiskit_gates():
 
 
 def translate_qiskit(source_circuit):
-    """Take in an abstract circuit, return an equivalent qiskit QuantumCircuit
-    instance
+    """Take in a Circuit, return an equivalent qiskit.QuantumCircuit
 
     Args:
-        source_circuit: quantum circuit in the abstract format.
+        source_circuit (Circuit): quantum circuit in the abstract format.
```
ValentinS4t1qbit, line 71: Maybe we should stop using "abstract" and simply say "Tangelo" format everywhere, I think its clearer.
You can just focus on changing what you see in your PR, dont worry about the other instances :)

tangelo/linq/translator/__init__.py
```diff
@@ -15,7 +15,7 @@
 from tangelo.helpers.utils import installed_backends
 
 from .translate_braket import translate_braket, get_braket_gates
-from .translate_qiskit import translate_qiskit, get_qiskit_gates
+from .translate_qiskit import translate_qiskit, get_qiskit_gates, translate_c_from_qiskit, translate_c_to_qiskit
```
ValentinS4t1qbit, line 18: Good for now !

Later on, we will stop importing those functions and instead make the user go through the "umbrella" function `translate_circuit` that takes source and target as argument, and calls the functions you have defined underneath. This way all translation functions are hidden under a single function that handles all cases.

ValentinS4t1qbit (changes requested):
Good stuff, a few small comments (some related to old docstrings that have not aged too well, oops). Should be quick to handle

tangelo/linq/translator/translate_qiskit.py
```diff
@@ -103,6 +115,48 @@ def translate_qiskit(source_circuit):
...
+        elif name in {"SWAP"}:
+            gates += [Gate(name, [qi[gate.qubits[0]].index, qi[gate.qubits[1]].index])]
+        elif name in {"CSWAP"}:
+            gates += [Gate(name, [qi[gate.qubits[1]].index, qi[gate.qubits[2]].index], control=qi[gate.qubits[0]].index)]
+        elif name in {"XX"}:
+            gates += [Gate(name, [qi[gate.qubits[0]].index, qi[gate.qubits[1]].index], parameter=gate.operation.params[0])]
+        elif name in {"MEASURE"}:
+            gates += [Gate(name, qi[gate.qubits[0]].index)]
+        else:
+            raise ValueError(f"Gate '{gate.name}' not supported in Tangelo")
```
ValentinS4t1qbit, line 157: All gates are supported in Tangelo (POTATO gate etc). We do not impose requirements for gate names, they only need to make sense when we translate from tangelo a well-defined framework. I don't believe this conditional statement is useful, it seems needlessly restrictive to me.

If tomorrow Qiskit defines a CARROT gate, not yet handled by tangelo, and a user wants to import that to use some functionalities to count gates in that circuit, remove redundant gates, export it back to Qiskit, I don't know why we would object to that. If the user tries to export or simulate in a format that does not support the gate, the dedicated functions will raise that error.

ValentinS4t1qbit (changes requested):

ValentinS4t1qbit (approved):

PR #204, opened 2022-08-26
Quantum Signal Processing time-evolution

JamesB-1qbit:
The Quantum Signal Processing algorithm for Hamiltonian Simulation as defined by arXiv:2002.11649.

- Uses pyqsp for the phases by default.
- More stable/faster function that uses interface oct2py with octave based QSPPACK also available

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -29,5 +29,5 @@ setuptools.setup(
     packages=setuptools.find_packages(),
     test_suite="tangelo",
     setup_requires=['h5py'],
-    install_requires=['h5py', 'bitarray', 'openfermion', 'openfermionpyscf', 'pyqsp']
+    install_requires=['h5py', 'bitarray', 'openfermion', 'openfermionpyscf']
 )
--- a/setup.py
+++ b/setup.py
@@ -29,5 +29,5 @@ setuptools.setup(
     packages=setuptools.find_packages(),
     test_suite="tangelo",
     setup_requires=['h5py'],
-    install_requires=['h5py', 'bitarray', 'openfermion', 'openfermionpyscf']
+    install_requires=['h5py', 'bitarray', 'openfermion', 'openfermionpyscf', 'pyqsp']
 )
--- a/tangelo/toolboxes/circuits/lcu.py
+++ b/tangelo/toolboxes/circuits/lcu.py
@@ -277,58 +277,7 @@ def get_oaa_lcu_circuit(qu_op: QubitOperator, control: Union[int, List[int]] = N
         Circuit: The circuit that implements the linear combination of unitaries for the qu_op
     """
 
-    if control is not None:
-        control_list = copy(control) if isinstance(control, list) else [control]
-    else:
-        control_list = []
-
-    unitaries = list()
-    vector = list()
-    max_qu_op = count_qubits(qu_op)
-    for term, coeff in qu_op.terms.items():
-        acoeff = np.abs(coeff)
-        vector += [acoeff]
-        unitaries += [QubitOperator(term, coeff / acoeff)]
-    vsum = sum(vector)
-
-    # Check that 1-norm is less than 2 and add and subtract terms proportional to the identity to obtain 1-norm equal to 2
-    if vsum > 2 + 1.e-10:
-        raise ValueError("Can not apply qu_op as its 1-norm is greater than 2")
-    vector += [max((2 - vsum)/2, 0), max((2 - vsum)/2, 0)]
-    unitaries += [QubitOperator(tuple(), 1), QubitOperator(tuple(), -1)]
-
-    # create U_{prep} from sqrt of coefficients
-    num_terms = len(vector)
-    vector = np.array(vector)
-    vector = np.sqrt(vector)/np.sqrt(2)
-    n_qubits = math.ceil(math.log2(num_terms))
-    newvec = np.zeros(2**n_qubits)
-    newvec[0:num_terms] = vector[:]
-    s = StateVector(newvec, order="lsq_first")
-    uprep = s.initializing_circuit()
-    uprep_qubits = list(range(max_qu_op, max_qu_op + n_qubits))
-    uprep.reindex_qubits(uprep_qubits)
-
-    # Generate U_{Select}
-    gate_list = list()
-    control_qubits = uprep_qubits + control_list
-    for j, unitary in enumerate(unitaries):
-        bs = bin(j).split('b')[-1]
-        state_binstr = "0" * (n_qubits - len(bs)) + bs
-        x_ladder = [Gate("X", q + max_qu_op) for q, i in enumerate(state_binstr) if i == "0"]
-        gate_list += x_ladder
-        for term, coeff in unitary.terms.items():
-            phasez = np.arctan2(-coeff.imag, coeff.real)
-            phasep = -phasez*2
-
-            if abs(phasez) > 1.e-12:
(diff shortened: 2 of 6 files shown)


setup.py
```diff
@@ -29,5 +29,5 @@ def install(package):
     packages=setuptools.find_packages(),
     test_suite="tangelo",
     setup_requires=['h5py'],
-    install_requires=['h5py', 'bitarray', 'openfermion', 'openfermionpyscf']
+    install_requires=['h5py', 'bitarray', 'openfermion', 'openfermionpyscf', 'pyqsp']
```
ValentinS4t1qbit, line 32: I wonder if we should require this in the install, considering its just a single feature, maintained by a third party. These other requirements here are more important or closer to python core.

Maybe we should have a try / error block that tells the user "this feature requires pyqsp, please install the pyqsp python package".

tangelo/toolboxes/circuits/lcu.py
```diff
@@ -20,6 +20,7 @@
 
 import math
 from typing import Union, Tuple, List
+from copy import copy
```
ValentinS4t1qbit, line 23: To make a shallow copy of a list just use the built-in copy method or slicing.
If your control lists contain integers (e.g immutable values) then the shallow copy is equivalent to a deepcopy.

```
l = [1,2,3]
l1 = l[:] # Full copy, via slicing
l2 = l.copy() # Shallow copy. Ok with integers etc

l.append('a')
l1.append('b')
l2.append('c')

# Each list is indeed separate / different
print(l, l1, l2)
```

tangelo/toolboxes/circuits/qsp.py
```diff
@@ -0,0 +1,272 @@
...
+        coef[i-1][0] = (-1)**(i-1)*jn(2*(i-1), tau)
+
+    coef[0] = coef[0]/2
+    phi1, _ = octave.QSP_solver(coef, 0, opts, nout=2)
+
+    coef = np.zeros((maxorder//2+1, 1), dtype=float, order='F')
+    for i in range(1, len(coef)+1):
+        coef[i-1][0] = (-1)**(i-1)*jn(2*i-1, tau)
+    phi2, _ = octave.QSP_solver(coef, 1, opts, nout=2)
+    return list(phi1.flatten()), list(phi2.flatten())
```
ValentinS4t1qbit, line 127: skip line before L127

tangelo/toolboxes/circuits/qsp.py
```diff
@@ -0,0 +1,272 @@
...
+    octave.addpath(folder)
+    opts = octave.struct("criteria", eps)
+    maxorder = math.ceil(1.4*tau+np.log(1e14))
+    if maxorder % 2 == 1:
+        maxorder -= 1
+
+    coef = np.zeros((maxorder//2 + 1, 1), dtype=float, order='F')
+    for i in range(1, len(coef)+1):
+        coef[i-1][0] = (-1)**(i-1)*jn(2*(i-1), tau)
+
```
ValentinS4t1qbit, line 119: Remove empty line to mimic the other similar code block

tangelo/toolboxes/circuits/qsp.py
```diff
@@ -0,0 +1,272 @@
...
+    if control is not None:
+        c_list += control if isinstance(control, list) else [control]
+
+    anc = m_qs[-1]+1
+    if with_z:
+        qubcirc = Circuit([Gate("CH", anc, control=c_list), Gate("CZ", anc, control=c_list)])
+    else:
+        qubcirc = Circuit([Gate("CH", anc, control=c_list)])
+
+    circ_angles = []
```
ValentinS4t1qbit, line 169: L170 defeats the point of L169

tangelo/toolboxes/circuits/qsp.py
```diff
@@ -0,0 +1,272 @@
...
+        Circuit: The QSP Hamiltonian simulation circuit with oblivious amplitude amplification to ensure very high success probability.
+    """
+
+    if control is not None:
+        control_list = copy(control) if isinstance(control, list) else [control]
+    else:
+        control_list = []
+
+    # If tau is negative, flip sign of tau and qu_op.
+    flip_tau_sign = False
```
ValentinS4t1qbit, line 220: set that variable to `flip_tau_sign = (tau < 0.)`.
Then I think L221-224 can be replaced by a duplicate of L268-270.

tangelo/toolboxes/circuits/qsp.py
```diff
@@ -0,0 +1,272 @@
...
+    circ += get_qsp_circuit_no_anc(cua, m_qs, anglesr, control=lcu_qs+control_list, with_z=False)
+    circ += Circuit([Gate("X", lcu_qs[0])])
+    # imaginary part i*sin(Ht)
+    circ += get_qsp_circuit_no_anc(cua, m_qs, anglesi, control=lcu_qs+control_list, with_z=False)
+    # -I to ensure probability is np.arcsin
+    circ += Circuit([Gate("X", lcu_qs[1]), Gate("CRZ", target=0, parameter=2*np.pi, control=lcu_qs+control_list), Gate("X", lcu_qs[1])])
+
+    circ += uprep.inverse()
+
+    if control is not None:
```
ValentinS4t1qbit, line 263: Out of curiosity: do we need the branching ?
What if we only use the first branch, and control=None is passed ?

It seems to make sense to me that control=None should naturally mean it's not a controlled gate and would return something no different than the non-controlled version of the gate, as expected.

tangelo/toolboxes/circuits/qsp.py
```diff
@@ -0,0 +1,272 @@
...
+"""Module to generate the circuits necessary to implement quantum signal processing (QSP) Hamiltonian Simulation
+
+Ref:
+[1]: Yulong Dong, Xiang Meng, K. Birgitta Whaley, Lin Lin, "Efficient phase-factor evaluation in quantum signal processing"
+2021, arXiv:2002.11649 (https://arxiv.org/abs/2002.11649)
+"""
+
+import math
+from typing import Union, List, Tuple
+from copy import copy
```
ValentinS4t1qbit, line 24: I dont think you need that (see other comments)

tangelo/toolboxes/circuits/qsp.py
```diff
@@ -0,0 +1,272 @@
...
+    Args:
+        qubit_list (List[int]): The qubits controlled by zero.
+        target (int): The target qubit.
+        control (List[int] or int): The qubits controlled by one.
+
+    Returns:
+        Circuit: The zero controlled cnot circuit with possible extra controls on 1.
+    """
+    x_ladder = [Gate("X", q) for q in qubit_list]
+    gates = x_ladder + [Gate("CX", target=target, control=qubit_list+control)] + x_ladder
```
ValentinS4t1qbit, line 142: Control can be an int or a union of int. What if control is an int here? Does this work ?

ValentinS4t1qbit (changes requested):
A few things that could be improved or cleaned up, but nothing big

tangelo/toolboxes/circuits/qsp.py
JamesB-1qbit: Yes, in general that is correct. Our current implementation forces users to specify controlled gates with a `C` in front of the gate name.  We can certainly revisit that if you want.

ValentinS4t1qbit: We can revisit that later if a user deems it a concern.

ValentinS4t1qbit (approved):
Thank you james !

Issue #206, opened 2022-09-08
[BUG] QEMIST Cloud credentials improperly checked when submitting jobs to IonQ Cloud


MPCoons:
# Issue: Bug Report

**Expected Behavior**
Running the attached Python script via "python minimum_example.py" is expected to

1. Create a SecondQuantizedMolecule for ground state H2 / STO-3G basis set.
2. Perform classical VQE to obtain the optimal QCC circuit for this system.
3. Connect with IonQ Cloud's API and submit the job to the simulator backend.
4. Return a histogram from the simulation.

**Current Behavior**
The job fails before connecting to IonQ Cloud and never submits the simulation to the simulator backend. Several errors related to the remote end disconnecting without response, a protocol error, and a connection error are reported. The full error message is attached to this report along with an example.

**Steps to Reproduce (minimal example)**
The job fails at Step 3 below.

1. Set the environment variable IONQ_APIKEY to its current value.
2. Activate the appropriate environment with the installation of tangelo-gc  v0.3.2.
3. Run the script with "python  minimum_example.py"

**Environment**
CentOS Linux release 7.8.2003 (Core)
The output of `conda list` is attached to this report. Briefly, tangelo-gc v0.3.2 and qemist-client v0.5.0 are installed.

**Possible Solution**
Commenting the line `from .qemist_cloud_connection import job_submit, job_status, job_cancel, job_result, job_estimate`  in `tangelo/linq/qpu_connection/__init__.py` fixed the issue.

[tangelo-gc_conda_env.txt](https://github.com/goodchemistryco/Tangelo/files/9530255/tangelo-gc_conda_env.txt)
[minimum_example_error.txt](https://github.com/goodchemistryco/Tangelo/files/9530256/minimum_example_error.txt)
[minimum_example_py.txt](https://github.com/goodchemistryco/Tangelo/files/9530257/minimum_example_py.txt)

ValentinS4t1qbit:
We indeed should not check or import anything QEMIST Cloud related unless the user explicitly use a feature that requires it.

A good way to make that clean is probably by encapsulating the qemist cloud "job" functions in `qemist_cloud_connection.py` in a `QCloudConnection` children class of `QPUConnection`, and hide the import / login checks inside the call to `__init__` (e.g unless the class is instantiated, users won't have any issue of imports. If the imports work at instantiation, there is no reason to check them for the other job methods).

[...]

PR #207, opened 2022-09-09
QPU connections doesn't need qemist_client

AlexandreF-1qbit:
Attempt at fixing #206. This bug happens when `qemist_client` is installed with Tangelo, and the QEMIST Cloud API is not responding. Even if `qemist_client` is installed in the environment, it should not be imported if we work with other QPU connections (e.g. IONQ).

One sketchy thing is the use of a `global qclient_util`. The only other way I thought to import the qemist_client functions in the `__init__` method was to make them attributes of the class. Please tell me if there is another way more efficient or elegant.


examples/qemist_cloud_hardware_experiments_braket.ipynb
```diff
@@ -111,7 +111,8 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "from tangelo.linq.qpu_connection import job_submit, job_status, job_cancel, job_result, job_estimate"
+    "from tangelo.linq.qpu_connection import QEMISTCloudConnection\n",
+    "qemist = QEMISTCloudConnection()"
```
ValentinS4t1qbit, line 115: Let's not call a class a name that has been used for our product.
`qcloud_connection` or `qemist_connection` may be better.

The term "connection" is rather well defined: https://docs.microsoft.com/en-us/sql/ado/reference/ado-api/connection-object-ado?view=sql-server-ver16

examples/overview_endtoend.ipynb
```diff
@@ -701,19 +701,21 @@
     "os.environ['QEMIST_AUTH_TOKEN'] = \"your_qemist_authentication_token\"\n",
     "\n",
     "# Estimate, submit and get the results of your job / quantum task through our wrappers\n",
-    "from tangelo.linq.qpu_connection import job_submit, job_status, job_cancel, job_result, job_estimate\n",
+    "from tangelo.linq.qpu_connection import QEMISTCloudConnection\n",
+    "\n",
+    "qemist = QEMISTCloudConnection()\n",
```
ValentinS4t1qbit, line 706: See other comment

ValentinS4t1qbit (changes requested):

ValentinS4t1qbit (approved):
Good stuff, thanks

PR #209, opened 2022-09-09
Add n_qubits option to MultiformOperator compress function

JamesB-1qbit:
The following code

```python3
from tangelo.molecule_library import mol_H4_sto3g
from tangelo.toolboxes.operators.taper_qubits import QubitTapering
from tangelo.toolboxes.qubit_mappings.mapping_transform import fermion_to_qubit_mapping

qu_op = fermion_to_qubit_mapping(mol_H4_sto3g.fermionic_hamiltonian, "JW", 8, 4, True, 0)
tapering = QubitTapering(qu_op, 8, 4, 0, "JW", True)
tapered_qu_op = tapering.z2_tapering(QubitOperator("X0 X1 X4 X5", 1), 8)
```

raised an IndexError on the last line.

This small fix adds an n_qubits option for `MultiformOperator.compress` and utilizes it in do_taper.


tangelo/toolboxes/operators/multiformoperator.py
```diff
@@ -230,7 +234,7 @@ def compress(self, abs_tol=None):
         else:
             super(QubitOperator, self).compress(abs_tol)
 
-        self._update()
+        self._update(n_qubits)
```
AlexandreF-1qbit, line 237: Good catch!

tangelo/toolboxes/operators/tests/test_taper_qubits.py
```diff
@@ -58,6 +58,19 @@ def test_z2taper_h2_jw_spinup_first(self):
...
 
+    def test_z2taper_h4_jw_spinup_first(self):
+        """Test that a small qubit operator will be tapered properly when using tapering from a larger operator.
+        In this case, a 6 qubit operator from an 8 qubit operator tapering."""
+
+        qu_op = load_operator("H4_JW_spinupfirst.data", data_directory=pwd_this_test+"/data", plain_text=True)
+
+        tapering = QubitTapering(qu_op, 8, 4, 0, "JW", True)
+
+        try:
```
AlexandreF-1qbit, line 69: What is the purpose of this test? Should we verify the result?

AlexandreF-1qbit (approved):

tangelo/toolboxes/operators/tests/test_taper_qubits.py
ValentinS4t1qbit: It would seem James is happy as long as it does not raise an error, but I would be happier if the test checked for the output value or one of its properties (number of qubits or terms we expected to be present etc).

ValentinS4t1qbit (approved):
Thank you james !

PR #210, opened 2022-09-12
even Trotter order > 2

JamesB-1qbit:
Adds functionality for even trotter order greater than 2.


ValentinS4t1qbit (approved):
Looks good to me :)

PR #213, opened 2022-09-15
Refactor simulator

JamesB-1qbit:
Makes each target it's own class with SimulatorBase as a parent class. The majority of the code is copied/pasted.


examples/linq/1.the_basics.ipynb
```diff
@@ -844,7 +844,7 @@
     "# Option2: Assume quantum circuit simulation was performed separately (by this package or different services)\n",
     "freqs, _ = sim.simulate(c)  # circuit c must be consistent with operator, regarding measurement along non-z axes\n",
     "term, coef = tuple(op.terms.items())[0]  # This yields ((0, 'Z'),), 1.0\n",
-    "expval2 = coef * Simulator.get_expectation_value_from_frequencies_oneterm(term, freqs)\n",
+    "expval2 = coef * sim.get_expectation_value_from_frequencies_oneterm(term, freqs)\n",
```
ValentinS4t1qbit, line 847: We've known this to be a horrendously long name for a while. I don't want to bother with it in this PR because I think we'd do it with a proper deprecation after Alexandre's PR has been merged. But I think `expval` or something is better than `get_expectation_value` and we should have something shorter than the one above.

ValentinS4t1qbit (changes requested):
I think we can improve a few things, overall I think this redesign is great and what we needed. Thank you for pushing this forward.

You probably have done it already, but think about how your design will be able to handle the multiplication of simulators / backends coming from Cirq, Qiskit, Qualms etc that we may want to support, as well as new simulator frameworks that simply have not been integrated to Tangelo at all.

tangelo/linq/target/target_cirq.py
```diff
@@ -0,0 +1,112 @@
...
+import numpy as np
+
+from tangelo.linq import Circuit
+from tangelo.linq.helpers.circuits.statevector import StateVector
+from tangelo.linq.simulator import SimulatorBase
+import tangelo.linq.translator as translator
+from tangelo.toolboxes.post_processing.histogram import Histogram
+
+
+class Cirq(SimulatorBase):
```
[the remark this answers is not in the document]
AlexandreF-1qbit: Or `CirqSim` would do also I think.

tangelo/linq/simulator.py
```diff
@@ -520,21 +372,66 @@ def _statevector_to_frequencies(self, statevector):
...
+
+    @property
+    def backend_info(self):
+        """A dictionary that includes {'noisy_simulation': True or False,
+                                       'statevector_available': True or False,
+                                       'statevector_order': 'lsq_first' or 'msq_first'"""
+        pass
+
+
+def Simulator(target=default_simulator, n_shots=None, noise_model=None, **kwargs) -> SimulatorBase:
```
AlexandreF-1qbit, line 402: Will we keep this wrapper function in the future?

tangelo/linq/simulator.py
```diff
@@ -520,21 +372,66 @@ def _statevector_to_frequencies(self, statevector):
...
+            expected from the target backend.
+
+    Returns:
+        BaseSimulator: The initialized target simulator that is a child class of BaseSimulator.
+    """
+    if target is None:
+        target = default_simulator
+    if isinstance(target, str):
+        from tangelo.linq.target import Cirq, Qulacs, Qiskit, QDK
+        target_dict = {"qiskit": Qiskit, "cirq": Cirq, "qdk": QDK, "qulacs": Qulacs}
```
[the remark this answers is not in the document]
JamesB-1qbit: Building target_dict outside of a function caused a circular import. `SimulatorBase` is now in its own file.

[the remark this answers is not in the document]
ValentinS4t1qbit: It is not a bad thing. To clarify: does that mean that moving the base class solved the issue and addressed the comment ? If yes, please resolve this conversation.

tangelo/linq/simulator.py
```diff
@@ -26,515 +26,35 @@
...
-        return state_binstr[::-1] if (self.statevector_order == "msq_first") else state_binstr
 
-    def __int_to_binstr_lsq(self, i, n_qubits):
-        """Convert an integer into a bit string of size n_qubits, in the
-        least-significant qubit order.
-        """
-        bs = bin(i).split('b')[-1]
-        state_binstr = "0" * (n_qubits - len(bs)) + bs
-        return state_binstr[::-1]
+def Simulator(target=default_simulator, n_shots=None, noise_model=None, **kwargs) -> SimulatorBase:
```
ValentinS4t1qbit, line 38: I did not realize, but this is a function wrapper that returns a specialized class that depends on the parameters passed to it.

Although it is a function, you kept the capital S, I assume in an effort to break nothing, but also because functionally this behaves like an object constructor (factory).

The other way to achieve this would be to make `Simulator` a class, and then return the appropriate object by throwing the body of your function into the `__new__` method and leaving `__init__` as `pass` (the latter effectively means that you simply will never instantiate `Simulate`, but always another class). A very little example:

```
# Pretend QiskitSimulator has been defined and inherits from SimulatorBase
class Simulator():
    def __new__(self, target):
        if target == 'qiskit':
            return QiskitSimulator()
    def __init__(self):
        pass

qiskit_sim = Simulator(target="qiskit")
print(type(qiskit_sim)) # this is indeed a QiskitSimulator object
```

My unease with your function wrapper:
- you claim it returns a `SimulatorBase` object and... well that's a bit misleading or unhelpful here. It returns a specialization of that class, which implements the method `simulate_circuit` so that's not quite the same.
- `Simulate` is a function masquerading as a class (name, return values), it is misleading as many users will rely on name conventions, but strictly adhering to name conventions will break retro-compatibility.

tangelo/linq/simulator.py
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""Simulator class, wrapping around the various simulators and abstracting their
+"""Simulator function, returns the various simulators that abstract their
```
ValentinS4t1qbit, line 15: Let's adjust the docstring here. Make Simulator a class again, its job is now to "return an instance of the class that supports the target backend"

tangelo/linq/simulator.py
ValentinS4t1qbit: We need a single entry-point to all our specialization of simulators, so we need a "Simulator" umbrella.

tangelo/helpers/utils.py
```diff
@@ -34,13 +35,9 @@ def __exit__(self, exc_type, exc_val, exc_tb):
...
 
 def is_package_installed(package_name):
-    try:
-        exec(f"import {package_name}")
-        # DEBUG print(f'{package_name:16s} :: found')
-        return True
-    except ModuleNotFoundError:
-        # DEBUG print(f'{package_name:16s} :: not found')
-        return False
+    """Check if module is installed without importing."""
```
ValentinS4t1qbit, line 38: Nice finding !

tangelo/linq/simulator.py
```diff
@@ -26,515 +26,35 @@
...
+
+    Returns:
+        BaseSimulator: The initialized target simulator that is a child class of BaseSimulator.
+    """
+    if target is None:
+        target = default_simulator
+    if isinstance(target, str):
+        simulator = target_dict[target](n_shots=n_shots, noise_model=noise_model)
+    else:
+        simulator = target(n_shots=n_shots, noise_model=noise_model, **kwargs)
```
ValentinS4t1qbit, line 58: What is the purpose of this branch ? (this suggests that regardless of your answer, we need comments).
If its purpose is redundant with the other branch, I suggest we remove it.

If its purpose brings out a new functionality / use cases, I need this functionality to be explained and tested. Unless I missed something, I don't think this has been tested. You have the option of leaving it out for another small PR, this way Simulator and Translator 2.0 can be merged a bit faster first.

tangelo/linq/simulator.py
```diff
@@ -26,515 +26,35 @@
...
+        n_shots (int): Number of shots if using a shot-based simulator.
+        noise_model: A noise model object assumed to be in the format
+            expected from the target backend.
+        kwargs: Other arguments that could be passed to a target. Examples are qubits_to_use for a QPU, transpiler optimization
+            level, error mitigation flags etc.
+
+    Returns:
+        BaseSimulator: The initialized target simulator that is a child class of BaseSimulator.
+    """
+    if target is None:
```
ValentinS4t1qbit, line 53: Not fond of this code. `target` can be 3 different types (None, a string, or.... anything and we'll try to shove it in L58?) and I'm unclear about how errors (e.g unsupported value for `target`) are currently handled in each of these scenarii.

tangelo/linq/simulator.py
```diff
@@ -26,515 +26,35 @@
 (e.g a number of shots is required).
 
 Some backends may only support a subset of the above. This information is
-contained in a separate data-structure.
+contained in backend_info.
```
ValentinS4t1qbit, line 29: backend_info does not exist here, so maybe specify "the backend_info variable under the linq module."

tangelo/linq/tests/test_simulator_noisy.py
```diff
@@ -65,6 +65,11 @@ def test_noisy_simulation_not_supported(self):
             Ensures that an error is returned if user attempts to run noisy simulation on a backend that does
             not support it as part of this package.
         """
+        backend_info = dict()
```
ValentinS4t1qbit, line 68: I am confused. Why is this information hardcoded into a test file ?
Is there a bug or circular dependency preventing us from using `backend_info` ?

ValentinS4t1qbit (changes requested):
I apologize, I reviewed a bit too quickly the first time and I think your Simulator wrapper function requires work.

I'm grateful for the opportunity to think about this design change, which brings out some subtleties we hadn't faced before.

tangelo/linq/simulator.py
JamesB-1qbit: This was tested in a previous version. It can be seen in (https://github.com/JamesB-1qbit/Tangelo/blob/refactor_simulator_noisy/examples/linq/qiskit_device.ipynb) where QiskitDevice is provided.

Added test for providing user provided subclass of SimulatorBase target.

The branch was needed to allow users to provide their own subclass of SimulatorBase. This is necessary such that algorithms such as VQESolver can be provided with the class TargetSimulator instead of "target" string.for the backend options.

tangelo/linq/simulator.py
```diff
@@ -26,515 +26,64 @@
...
-backend_info["cirq"] = {"statevector_available": True, "statevector_order": "lsq_first", "noisy_simulation": True}
-backend_info["qdk"] = {"statevector_available": False, "statevector_order": None, "noisy_simulation": False}
+class Simulator(SimulatorBase):
 
-
-class Simulator:
-
-    def __init__(self, target=default_simulator, n_shots=None, noise_model=None):
-        """Instantiate Simulator object.
+    def __new__(self, target=default_simulator, n_shots=None, noise_model=None, **kwargs):
```
ValentinS4t1qbit, line 44: Take those wonderful type checks from `__init__` and use them here if you think they're useful.
Leave it alone and delete the typing imports if you think otherwise (NB: you were not consistent in your type checks in your own code file).

tangelo/linq/simulator.py
```diff
@@ -26,515 +26,64 @@
...
+        """Return requested target simulator class
```
ValentinS4t1qbit, line 45: You are not returning a class, you are returning an object (e.g an instance of the target simulator class).
The class is the blueprint giving the recipe to build such objects.

tangelo/linq/simulator.py
```diff
@@ -26,515 +26,64 @@
...
-        if are_coefficients_real:
-            if self._noise_model or not self.statevector_available \
-                    or state_prep_circuit.is_mixed_state or state_prep_circuit.size == 0:
-                return self._get_expectation_value_from_frequencies(qubit_operator, state_prep_circuit, initial_statevector=initial_statevector)
-            elif self.statevector_available:
-                return self._get_expectation_value_from_statevector(qubit_operator, state_prep_circuit, initial_statevector=initial_statevector)
-
-        # Else, separate the operator into 2 hermitian operators, use linearity and call this function twice
+        # Return new instance of target class
+        if target is None:
```
ValentinS4t1qbit, line 58: I don't think L58-59 should exist anymore.

Your method signature handles a default value, neither your documentation nor your error message below mentions `None` as a supported value.

With your default value, users would have to go out of their way to stipulate `target=None` to be in this situation, which seems non-sensical to me.

tangelo/linq/simulator.py
```diff
@@ -26,515 +26,64 @@
...
 
         Args:
-            target (str): One of the available target backends (quantum or
-                classical). The default simulator is qulacs if found in user's
-                environment, otherwise cirq.
+            target (string or Type[SimulatorBase]): String can be "qiskit", "cirq", "qdk" or "qulacs". Can also provide
+                a subclass of SimulatorBase.
             n_shots (int): Number of shots if using a shot-based simulator.
-            noise_model: A noise model object assumed to be in the format
+            noise_model (NoiseModel): A noise model object assumed to be in the format
```
ValentinS4t1qbit, line 51: If it's supposed to be in the format supported by the target backend, then we don't know what type this argument is. It is whatever the user / builtin implementation handles.

tangelo/linq/simulator.py
```diff
@@ -26,515 +26,64 @@
...
-import os
-import math
-from collections import Counter
+from typing import Union, Tuple, Type
 
-import numpy as np
-from scipy import stats
-from bitarray import bitarray
-from openfermion.ops import QubitOperator
+import numpy
```
ValentinS4t1qbit, line 34: I don't believe your code needs this anymore. Does your IDE highlight unused packages? I think this is not the first time we come across that.

ValentinS4t1qbit (changes requested):
A number of improvements to make. I want to recognize that this PR is getting a lot of comments because:

- you are introducing several non-trivial **badass** features
- you are introducing patterns we haven't used before, therefore learning and challenging us and yourself.

As far as I'm concerned, this is an amazing PR. Please keep up the good work and take us to the finish line.

tangelo/linq/simulator.py
ValentinS4t1qbit: Could we add a test that demonstrate that indeed an existing algorithm can work with a custom backend ? Even if the simulator is a dummy and the result somewhat non-sensical, et predictable (e.g maybe the simulator returns a hardcoded statevector, maybe the zero state, which makes the optimization process trivial and yield a particular energy)

tangelo/linq/simulator.py
```diff
@@ -26,515 +26,64 @@
...
-
-        Args:
-            qubit_operator(openfermion-style QubitOperator class): a qubit
-                operator.
-            state_prep_circuit: an abstract circuit used for state preparation.
-
-        Returns:
-            complex: The expectation value of this operator with regards to the
-                state preparation.
+            target (string or Type[SimulatorBase]): String can be "qiskit", "cirq", "qdk" or "qulacs". Can also provide
```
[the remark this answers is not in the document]
JamesB-1qbit: Unfortunately my IDE uses `__init__` to provide docstring info for initialization and not `__new__` so I changed the process to run everything through `__init__`. `__new__` is no longer there.  I also added a description to the `Simulator` Class. So in my IDE (which is VS Code).

mouse over on
```python
Simulator
```
returns 
```
Class that when initialized becomes an instance of a target subclass of SimulatorBase. All available targets can be
found in tangelo.linq.target. Can also be used to access static methods of SimulatorBase without initializing.
```

mouseover on
```python
Simulator()
```
returns
```
Initialize requested target simulator object.

   Args:
       target (string or Type[SimulatorBase]): String can be "qiskit", "cirq", "qdk" or "qulacs". Can also provide
           a subclass of SimulatorBase.
       n_shots (int): Number of shots if using a shot-based simulator.
       noise_model: A noise model object assumed to be in the format expected from the target backend.
       kwargs: Other arguments that could be passed to a target. Examples are qubits_to_use for a QPU, transpiler
           optimization level, error mitigation flags etc.
```

If you don't leave a docstring on `__init__`, it uses the SimulatorBase __init__ which is.
```
Instantiate Simulator object.

   Args:
       n_shots (int): Number of shots if using a shot-based simulator.
       noise_model: A noise model object assumed to be in the format
           expected from the target backend.
```
which is wrong for the Simulator class.

tangelo/linq/simulator.py
```diff
@@ -26,515 +26,64 @@
...
-import tangelo.linq.translator as translator
 
 
-# Data-structure showing what functionalities are supported by the backend, in this package
-backend_info = dict()
-backend_info["qiskit"] = {"statevector_available": True, "statevector_order": "msq_first", "noisy_simulation": True}
-backend_info["qulacs"] = {"statevector_available": True, "statevector_order": "msq_first", "noisy_simulation": True}
-backend_info["cirq"] = {"statevector_available": True, "statevector_order": "lsq_first", "noisy_simulation": True}
-backend_info["qdk"] = {"statevector_available": False, "statevector_order": None, "noisy_simulation": False}
+class Simulator(SimulatorBase):
```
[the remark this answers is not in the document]
JamesB-1qbit: This is so it has access to static methods of SimulatorBase

tangelo/linq/simulator.py
JamesB-1qbit: Unfortunately, all our algorithms `VQESolver, ADAPTSolver,...` that use `Simulator` send `None` as the default `target`. I could change that behaviour in this PR but that adds many changed files to this PR.

tangelo/linq/simulator.py
JamesB-1qbit: numpy was used for a typing but I realized that `simulate_circuit` is really a local method and the typing can be provided by `simulate`.

tangelo/linq/simulator.py
ValentinS4t1qbit: Thank you ! I think this could be a separate PR then.
For now, if you are supporting `None` as a value for target, your current documentation and error message should however be consistent.

AlexandreF-1qbit (approved):
I reviewed only the big picture, and I like the modularity it will bring. Will looking at `TrueFalseSimulator`, it seems easy to implement another backend rapidly.

examples/linq/2.qpu_connection.ipynb
```diff
@@ -767,24 +767,24 @@
...
+    "from tangelo.linq.simulator_base import SimulatorBase\n",
     "from tangelo.linq.helpers.circuits.measurement_basis import pauli_string_to_of\n",
     "\n",
     "# Rescale counts to obtain frequency histogram\n",
     "freqs = {k:v/100 for k,v in counts.items()}\n",
     "print(f\"Frequencies: {freqs}\")\n",
     "\n",
     "# Compute expectation value for ZZ operator\n",
-    "exp_ZZ = Simulator.get_expectation_value_from_frequencies_oneterm(pauli_string_to_of(\"ZZ\"), freqs)\n",
+    "exp_ZZ = SimulatorBase.get_expectation_value_from_frequencies_oneterm(pauli_string_to_of(\"ZZ\"), freqs)\n",
```
ValentinS4t1qbit, line 778: Frustratingly obvious at this point (wish we realized it sooner): why are we importing a class if our goal is to run a static function that does not need anything from that class?

We should define `get_expectation_value_from_frequencies_oneterm` as a standalone function and import it directly (how about we import it with a smaller name btw? Like`expval_from_freqs_oneterm` or something, to use it anywhere a target simulator class is not needed.

Then we define a method in `SimulatorBase` that basically calls that function. As a result, the standalone function is available, but also usable through the children of the `SimulatorBase` classes.

We thus have no reasons to import `SimulatorBase`: we just import the function directly or we use an existing target simulator object returned by the `Simulator` function.

tangelo/linq/simulator_base.py
```diff
@@ -0,0 +1,400 @@
...
+
+    def __init__(self, n_shots=None, noise_model=None):
+        """Instantiate Simulator object.
+
+        Args:
+            n_shots (int): Number of shots if using a shot-based simulator.
+            noise_model: A noise model object assumed to be in the format
+                expected from the target backend.
+        """
+        self._source = "abstract"
```
ValentinS4t1qbit, line 56: Do a quick check: I think this attribute is unused. It had to do with translation, and was used in case we pushed non-tangelo circuit to this class. I think it is now irrelevant.

tangelo/toolboxes/post_processing/histogram.py
```diff
@@ -18,7 +18,7 @@
 
 from collections import Counter
 
-from tangelo.linq import Simulator
+from tangelo.linq.simulator_base import SimulatorBase
```
ValentinS4t1qbit, line 21: Let's directly import the function and not that class. (there are probably a few other files in your PR where this happens)

ValentinS4t1qbit (changes requested):
I swear we are getting there.

We are currently importing a class (`SimulatorBase`) but in reality we just need a function that does not care at all about anything class-related. Instead let's define and import a standalone function. The `SimulatorBase` can then use that function to implement a method with the same name, and that method can then be automatically leveraged by the children classes, as it currently does.

This way we have no loss of functionality, and the imports are straightforward. This also means we never need to instantiate `SimulatorBase`, which is intended to just be a base class for target simulators. Everything seems to fit into place with a clear role, which is a sign of a good design.

ValentinS4t1qbit (approved):
Great job James, glad to see this now part of our code !

Issue #211, opened 2022-09-12
[BUG] The type of object returned by the `do_jkmn_transform(vector)` function in `statevector_mapping.py` is incorrect.


MPCoons:
# Issue: Bug Report

**Expected Behavior**
The `do_jkmn_transform(vector)` function is supposed to return a `numpy array of int`. 

**Current Behavior**
Instead, the `do_jkmn_transform(vector)` function returns a `list of int`. See the `prep_jkmn_vector(vector)` function in `jkmn.py` -- the output is a `list`.

QCC jobs fail if the JKMN encoding is used. The failure arises when initializing the qubit mean field variational parameters via the `init_qmf_from_hf` function: this function requires a `numpy array of int` from the `statevector_mapping` module.

**Steps to Reproduce (minimal example)**
Rename the attached `test_qcc_jkmn_py.txt` file to `test_qcc_jkmn.py`.
Then run `python test_qcc_jkmn.py`.
The resulting error message is attached as `test_qcc_jkmn_error.txt`

**Environment**
CentOS Linux release 7.8.2003 (Core)
Tangelo-gc v0.3.2 is installed.

**Possible Solution**
Change the output of the `do_jkmn_transform(vector)` function in `statevector_mapping.py` to `np.array(prep_jkmn_vector(vector))`. Alternatively, the output of `prep_jkmn_vector(vector)` can be changed to `x_vector` instead of `list(x_vector)`.

[test_qcc_jkmn_error.txt](https://github.com/goodchemistryco/Tangelo/files/9551634/test_qcc_jkmn_error.txt)
[test_qcc_jkmn_py.txt](https://github.com/goodchemistryco/Tangelo/files/9551635/test_qcc_jkmn_py.txt)

ValentinS4t1qbit:
Resolved with #215

PR #215, opened 2022-09-15
JKMN vector prep should return an array of int

AlexandreF-1qbit:
Related to issue #211. Implementation of QCC is using the numpy broadcasting function, thus raising an error when selecting the JKMN mapping (the vector is a list).


JamesB-1qbit (commented):
I believe I was not careful when changing the scbk version as it returns a numpy array of float. Could you also fix that also?

AlexandreF-1qbit:
Sure!

ValentinS4t1qbit (approved):

ValentinS4t1qbit (approved):

PR #217, opened 2022-09-20
Histogram class

AlexandreF-1qbit:
Histogram class provides functions to manipulate results from an experiment. The two highlights are the `aggregate_histograms` (or `__add__` method) and `post_select` functions. The former one combines many histograms into a single one. The later function removes data according to the outcome of a (or many) qubit(s).

### DIFF
--- a/tangelo/toolboxes/post_processing/histogram.py
+++ b/tangelo/toolboxes/post_processing/histogram.py
@@ -25,18 +25,17 @@ from tangelo.toolboxes.post_processing.bootstrapping import get_resampled_freque
 class Histogram:
     """Class to provide useful tools helping redundant tasks when analyzing
     data from an experiment. The expected data input is an histogram of
-    bistrings ("010..."), where 0/1 correspond to the |0>/|1> measured state in
-    the computational basis. The outcomes can refer to either the number of
-    replicas (where the summation across the bistrings equals the number of
-    shots), or the probability (where the summation equals to 1). In the former
-    case, the number of shots must be provided.
+    bistrings ("010..."), where 0 (resp. 1) correspond to the |0> (resp. |1>)
+    measured state in the computational basis. The outcomes can refer to either
+    the shot counts for each computational basis, or the corresponding
+    probability.
 
     The internal representation is kept as an histogram of shots. Normalized
     quantities are accessible via the Histogram "frequencies" properties.
 
     Args:
-        outcomes (dict): Results in the format of bistring: outcome, where
-            outcome can be an int (number of replicas) or a float (probability).
+        outcomes (dict of string: int or float): Results in the format of bistring:
+        outcome, where outcome can be a number of shots a probability.
         n_shots (int): Self-explanatory. If it is greater than 0, the class
             considers that probabilities are provided.
         msq_first (bool): Bit ordering. For example, 011 (msq_first) = 110
@@ -60,6 +59,8 @@ class Histogram:
                 raise ValueError(f"Histogram frequencies not summing very close to 1. (sum = {sum_values}).")
 
             outcomes = {k: round(v*n_shots) for k, v in outcomes.items()}
+        elif n_shots < 0:
+            raise ValueError(f"The number of shots provided ({n_shots}) must be a positive value.")
 
         self.counts = outcomes.copy()
 
@@ -112,9 +113,26 @@ class Histogram:
         """
         return {bistring: counts/self.n_shots for bistring, counts in self.counts.items()}
 
-    def post_select(self,  expected_outcomes):
-        """Post selection is done with the post_select function (see the
-        relevant documentation for more details). This method change the Histogram
+    def remove_bits(self, *indices):
+        """Method to remove bits on given indices.
+
+        Args:
+            indices (variable number of int): Qubit indices to remove.
+        """
+        new_counts = dict()
+        for bitstring, counts in self.counts.items():
+            new_bistring = "".join([bitstring[qubit_i] for qubit_i in range(len(bitstring)) if qubit_i not in indices])
+
+            if new_bistring in new_counts:
+                new_counts[new_bistring] += counts
+            else:
+                new_counts[new_bistring] = counts
+
+        self.counts = new_counts
+
+    def post_select(self, expected_outcomes):
+        """Method to apply post selection on Histogram data, based on a post
+        selection function in this module. This method change the Histogram
         object inplace.
 
         Args:
@@ -123,8 +141,15 @@ class Histogram:
                 based on the first qubit with the 1 state measured. This argument
                 can also filter many qubits.
         """
-        new_hist = post_select(self, expected_outcomes)
-        self.__dict__ = new_hist.__dict__
+        def f_post_select(bitstring):
+            for qubit_i, expected_bit in expected_outcomes.items():
+                if bitstring[qubit_i] != expected_bit:
+                    return False
+            return True
+
+        new_hist = filter_hist(self, f_post_select)
+        self.counts = new_hist.counts
+        self.remove_bits(*list(expected_outcomes.keys()))
 
     def resample(self, n_shots):
         """Post selection is done with the tangelo.toolboxes.post_processing.boostrapping.get_resampled_frequencies
@@ -141,8 +166,8 @@ class Histogram:
 
     def get_expectation_value(self, term, coeff=1.):
         """Output the expectation value for qubit operator term. The histogram
-        data is expected to be results from a circuit with the proper qubit
-        rotations.
+        data is expected to have been acquired in a compatible measurement
+        basis.
 
         Args:
             term(openfermion-style QubitOperator object): a qubit operator, with
@@ -156,7 +181,8 @@ class Histogram:
 
 
 def aggregate_histograms(*hists):
-    """Aggregate all input Histogram objects together.
+    """Return a Histogram object formed from data aggregated from all input
+    Histogram objects.
 
     Args:
         hists (variable number of Histogram objects): the input Histogram objects.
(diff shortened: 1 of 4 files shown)


JamesB-1qbit (approved):

tangelo/toolboxes/post_processing/histogram.py
```diff
@@ -0,0 +1,209 @@
...
+    def frequencies(self):
+        """The frequencies are normalized counts vs the number of shots.
+
+        Returns:
+            dict: Frequencies in a {bistring: probability, ...} format.
+        """
+        return {bistring: counts/self.n_shots for bistring, counts in self.counts.items()}
+
+    def post_select(self,  expected_outcomes):
+        """Post selection is done with the post_select function (see the
```
ValentinS4t1qbit, line 116: How about we say "Method to apply post selection on Histogram data, based on a post selection function in this module." for the first part

tangelo/toolboxes/post_processing/histogram.py
```diff
@@ -0,0 +1,209 @@
...
+        n_shots (int): Self-explanatory. If it is greater than 0, the class
+            considers that probabilities are provided.
+        msq_first (bool): Bit ordering. For example, 011 (msq_first) = 110
+            (lsq_first). This depends on the hardware provider convention.
+        epsilon (float): When probabilities are provided, this parameter is used
+            to check if the sum is within [1-epsilon, 1+epsilon]. Default is
+            1e-2.
+    """
+
+    def __init__(self, outcomes, n_shots=0, msq_first=False, epsilon=1e-2):
```
ValentinS4t1qbit, line 49: What makes more sense to you: default value for shots=0 or `None` ?

Regardless of what you pick, a number of operations here will return strange things if n_shots is not a strictly positive number, we may want to to account for that with a warning message, errors, or something like that.

ValentinS4t1qbit (changes requested):
Solid and exciting work: I think some aspects of the Histogram class can be improved a bit. I encourage you to think about how you'd go to support what we have done in the recent hardware experiment (post-selection on bit strings with qubits measured in a specific state) but also how users would handle other types of post-selection (e.g find all entries whose state preserves spin or charge, stuff like that).

It is perfectly ok to stop at the former in this PR if it's unclear how we would facilitate the latter (everything at the end can be done with a dictionary comprehension and a filter/predicate function that returns True for the bit strings that match what we want and False for the others).

tangelo/toolboxes/post_processing/histogram.py
```diff
@@ -0,0 +1,209 @@
...
+        return len(list(self.counts.keys())[0])
+
+    @property
+    def frequencies(self):
+        """The frequencies are normalized counts vs the number of shots.
+
+        Returns:
+            dict: Frequencies in a {bistring: probability, ...} format.
+        """
+        return {bistring: counts/self.n_shots for bistring, counts in self.counts.items()}
```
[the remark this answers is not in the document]
AlexandreF-1qbit: n_shots is not saved as an attribute, it is a property returning `sum(self.counts.values())`.

tangelo/toolboxes/post_processing/histogram.py
```diff
@@ -0,0 +1,209 @@
...
+            raise ValueError(f"Input histograms have different bitstring lengths ({lengths_bitstrings})")
+
+        # Aggregate histograms.
+        total_counter = sum([Counter({k: v for k, v in h.counts.items()}) for h in hists], Counter())
+
+        return Histogram(dict(total_counter))
+
+
+def post_select(hist, expected_outcomes):
+    """Post selection function to select data when a supplementary circuit is
```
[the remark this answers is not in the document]
AlexandreF-1qbit: Good suggestion, I'll try to come up with something shortly.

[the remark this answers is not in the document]
AlexandreF-1qbit: I came up with something like that
```python
def filter_hist(hist, function, *args, **kwargs):
    new_counts = {bitstring: counts for bitstring, counts in hist.counts.items() if function(bitstring, *args, **kwargs)}
    return Histogram(new_counts, msq_first=False)
```

Users should be able to provide whatever function to apply on the bitstrings (summation, checking only one qubits, etc.). For example, I converted the previous `post_select` method to:
```python
    def post_select(self, expected_outcomes):
        def f_post_select(bitstring):
            for qubit_i, expected_bit in expected_outcomes.items():
                if bitstring[qubit_i] != expected_bit:
                    return False
            return True

        new_hist = filter_hist(self, f_post_select)
        self.counts = new_hist.counts
        self.remove_bits(*list(expected_outcomes.keys()))
```

tangelo/toolboxes/post_processing/histogram.py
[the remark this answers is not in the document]
ValentinS4t1qbit: Here im worried about `counts/self.n_shots`

tangelo/toolboxes/post_processing/histogram.py
```diff
@@ -0,0 +1,225 @@
...
+        """Method to remove bits on given indices.
+
+        Args:
+            indices (variable number of int): Qubit indices to remove.
+        """
+        new_counts = dict()
+        for bitstring, counts in self.counts.items():
+            new_bistring = "".join([bitstring[qubit_i] for qubit_i in range(len(bitstring)) if qubit_i not in indices])
+
+            if new_bistring in new_counts:
```
ValentinS4t1qbit, line 126: On a un moyen facile de faire ca, quand on ne sait pas si une clef est présente dans un dictionnaire, avec la méthode get, qui procure une valeur par défaut si la clé n'est pas trouvée.

L126-129 devient

`new_counts[new_bitstring] = new_counts.get(new_bitstring, 0) + counts`

tangelo/toolboxes/post_processing/histogram.py
```diff
@@ -0,0 +1,225 @@
...
+            raise ValueError(f"Input histograms have different bitstring lengths ({lengths_bitstrings})")
+
+        # Aggregate histograms.
+        total_counter = sum([Counter({k: v for k, v in h.counts.items()}) for h in hists], Counter())
+
+        return Histogram(dict(total_counter))
+
+
+def filter_hist(hist, function, *args, **kwargs):
+    """Filter selection function to consider bistrings in respect to a boolean
```
ValentinS4t1qbit, line 211: Btw such function is called a predicate.

Here's how in practice people have been doing it with list comprehension, and in older python with `filter`: https://www.geeksforgeeks.org/filter-python-list-by-predicate-in-python/

tangelo/toolboxes/post_processing/histogram.py
```diff
@@ -0,0 +1,225 @@
...
+
+            if new_bistring in new_counts:
+                new_counts[new_bistring] += counts
+            else:
+                new_counts[new_bistring] = counts
+
+        self.counts = new_counts
+
+    def post_select(self, expected_outcomes):
+        """Method to apply post selection on Histogram data, based on a post
```
ValentinS4t1qbit, line 134: Post-selection may be unambiguous for you, but for me it is somewhat. Should we include a link or mention this explicitly done on desired outcomes for specific qubit indices ?

tangelo/toolboxes/post_processing/histogram.py
```diff
@@ -0,0 +1,225 @@
...
+        selection function in this module. This method change the Histogram
+        object inplace.
+
+        Args:
+            expected_outcomes (dict): Wanted outcomes on certain qubit indices and
```
ValentinS4t1qbit, line 139: wanted => desired.

1 state => |1> state.

tangelo/toolboxes/post_processing/histogram.py
```diff
@@ -0,0 +1,225 @@
...
+                their expected state. For example, {0: "1"} would filter results
+                based on the first qubit with the 1 state measured. This argument
```
ValentinS4t1qbit, line 141: "This argument can also filter many qubits" is something I don't understand very well. If you just mean that we can specify more than one qubit index in the dictionary, well, that kind of sounds obvious from the rest of your description. If you do need it though, may as well give an example with a dictionary of size 2 or something instead of the one you have right now.

tangelo/toolboxes/post_processing/histogram.py
```diff
@@ -0,0 +1,225 @@
...
+    @property
+    def frequencies(self):
+        """The frequencies are normalized counts vs the number of shots.
+
+        Returns:
+            dict: Frequencies in a {bistring: probability, ...} format.
+        """
+        return {bistring: counts/self.n_shots for bistring, counts in self.counts.items()}
+
+    def remove_bits(self, *indices):
```
ValentinS4t1qbit, line 116: Is this a function you want to expose to your users? If not, then put a `_` in front of the method name.
I let you decide whether this method is actually necessary, or if the body of the function could have simply been written in the `post_select` method, as it seems to be only connected to that.

Your name and the docstring / argument description seems a bit confusing.
You're not removing bits, you're effectively removing bitstrings entries from your histogram (and as result, their count is added to/as another entry).

tangelo/toolboxes/post_processing/histogram.py
```diff
@@ -0,0 +1,225 @@
...
+                if bitstring[qubit_i] != expected_bit:
+                    return False
+            return True
+
+        new_hist = filter_hist(self, f_post_select)
+        self.counts = new_hist.counts
+        self.remove_bits(*list(expected_outcomes.keys()))
+
+    def resample(self, n_shots):
+        """Post selection is done with the tangelo.toolboxes.post_processing.boostrapping.get_resampled_frequencies
```
ValentinS4t1qbit, line 155: I'd avoid naming the module etc because we need to change this if the function is renamed or moved.
You already have an import in the file, the user is a big boy: they can figure it out.

Further more this docstring is incorrect: this is not about post selection. Simply state that you are generating a new histogram with n_shots counts, based on the distribution given by the input Histogram object.

ValentinS4t1qbit (approved):
I think the design is good, the comments are just a collection of small things (docs, small expressions)

tangelo/toolboxes/post_processing/histogram.py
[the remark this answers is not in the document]
AlexandreF-1qbit: If `sum(self.counts.values()) = 0`, it means that there is no data in the Histogram or some results have negative values. We can put checks on that, but is it necessary with the doc strings?

tangelo/toolboxes/post_processing/histogram.py
AlexandreF-1qbit: So much cleaner ^^'

tangelo/toolboxes/post_processing/histogram.py
AlexandreF-1qbit: I do want to expose this function to users. It can be part of more complex filtering procedure.

tangelo/toolboxes/post_processing/histogram.py
```diff
@@ -0,0 +1,224 @@
...
+        new_counts = dict()
+        for bitstring, counts in self.counts.items():
+            new_bitstring = "".join([bitstring[qubit_i] for qubit_i in range(len(bitstring)) if qubit_i not in indices])
+            new_counts[new_bitstring] = new_counts.get(new_bitstring, 0) + counts
+
+        self.counts = new_counts
+
+    def post_select(self, expected_outcomes):
+        """Method to apply post selection on Histogram data, based on a post
+        selection function in this module. This method change the Histogram
```
JamesB-1qbit, line 132: change -> changes

tangelo/toolboxes/post_processing/histogram.py
```diff
@@ -0,0 +1,224 @@
...
+class Histogram:
+    """Class to provide useful tools helping redundant tasks when analyzing
+    data from an experiment. The expected data input is an histogram of
+    bistrings ("010..."), where 0 (resp. 1) correspond to the |0> (resp. |1>)
+    measured state in the computational basis. The outcomes can refer to either
+    the shot counts for each computational basis, or the corresponding
+    probability.
+
+    The internal representation is kept as an histogram of shots. Normalized
+    quantities are accessible via the Histogram "frequencies" properties.
```
JamesB-1qbit, line 34: Should mention somewhere that `lsq_first` is the internal storage format.

tangelo/toolboxes/post_processing/histogram.py
```diff
@@ -0,0 +1,224 @@
...
+        return sum(self.counts.values())
+
+    @property
+    def n_qubits(self):
+        """The length of the bistrings represents the number of qubits."
+
+        Returns:
+            int: Self-explanatory.
+        """
+        return len(list(self.counts.keys())[0])
```
JamesB-1qbit, line 105: This could also be `len(next(iter(self.counts)))`.

JamesB-1qbit (changes requested):
A copy function for Histogram may be nice. 
You seem to use both `bistring` and `bitstring`. Is this desired?

JamesB-1qbit (approved):

ValentinS4t1qbit (approved):

PR #218, opened 2022-09-21
Remove the n_min_orbitals attribute

AlexandreF-1qbit:
This attributes was meant to store the number of orbitals when using a minimal basis set. Changing the default basis cause this attribute to output wrong values, thus causing inconsistencies. This attribute is not used in our code base.


JamesB-1qbit (approved):

ValentinS4t1qbit (approved):
Thank you for keeping our codebase clean and clear, Alexandre.

PR #219, opened 2022-09-21
Change JKMN Majorana definition so mapped Hamiltonian is real

JamesB-1qbit:
The previous implementation resulted in a Qubit Hamiltonian that had real and imaginary parts. This change ensures that a real FermionOperator will be mapped to a real QubitOperator. (i.e. even number of Y elements with real coefficient or odd number of Y elements terms with imaginary coefficient)


AlexandreF-1qbit (approved):

ValentinS4t1qbit (approved):
Thank you James !

PR #220, opened 2022-09-23
Compute energy from RDMs with FermionOperator

AlexandreF-1qbit:
Before, we had to create a `SecondQuantizedMolecule` to compute the energy from RDMs. When dealing with `frozen_orbitals`, or different `mo_coeff`, this is adding unnecessary code lines. The coefficients have already been computed  and embedded into the Fermion Hamiltonian.


tangelo/toolboxes/molecular_computation/molecule.py
```diff
@@ -65,6 +65,40 @@ def molecule_to_secondquantizedmolecule(mol, basis_set="sto-3g", frozen_orbitals
...
+
+    # Loop through coefficients.
+    for p in range(n_mos):
+        for q in range(n_mos):
+            # Populate 1-body integrals.
+            one_body_integrals[p, q] = one_body_coefficients[2 * p, 2 * q]
+            # Continue looping to prepare 2-body integrals.
+            for r in range(n_mos):
+                for s in range(n_mos):
+                    two_body_integrals[p, q, r, s] = two_body_coefficients[2 * p, 2 * q + 1, 2 * r + 1, 2 * s]
```
ValentinS4t1qbit, line 97: For readability, I suggest removing spaces around the multiplication signs.
Readability directly result from varying the number of spaces between characters: putting the same amount of spaces everywhere is not much different than putting spaces nowhere.

ex: `two_body_coefficients[2*p, 2*q + 1, 2*r + 1, 2*s]`

tangelo/toolboxes/molecular_computation/tests/test_molecule.py
```diff
@@ -198,6 +214,27 @@ def test_ecp(self):
...
+
+        # Should work.
+        dummy_mo_coeff = np.array([
+            [0., 1.],
+            [2., 3.]
+        ])
+        molecule.mo_coeff = dummy_mo_coeff
+
+        # Should raise an AssertionError.
+        bad_dummy_mo_coeff = np.array([
```
ValentinS4t1qbit, line 230: In efficient bumpy, this is `np.arange(9).reshape((3,3))`.
If we don't care about the numbers, why not just np.ones((3,3)) ? Stay simple.

tangelo/toolboxes/molecular_computation/tests/test_rdms.py
```diff
@@ -0,0 +1,56 @@
...
+        but from a fermionic operator instead.
+        """
+        ferm_op_of = load_operator("H2_ferm_op.data", data_directory=pwd_this_test+"/data", plain_text=True)
+        ferm_op = FermionOperator()
+        ferm_op.__dict__ = ferm_op_of.__dict__
+
+        rdm1 = [[ 1.97453997e+00, -7.05987336e-17],
+                [-7.05987336e-17,  2.54600303e-02]]
+
+        rdm2 = [
```
ValentinS4t1qbit, line 40: May I suggest 4 entries 4 lines instead of two to improve readability and use less lines?

tangelo/toolboxes/operators/tests/test_operators.py
```diff
@@ -39,6 +41,51 @@ def test_qubitop_to_qubitham(self):
...
+        dummy_ferm_op += FermionOperator("0^ 0^ 0 0", 7.)
+        dummy_ferm_op += FermionOperator("1^ 1^ 1 1", 8.)
+        dummy_ferm_op += FermionOperator("0^ 1^ 0 1", 9.)
+
+        ref_cte = 1.
+        ref_one_body = np.array([
+            [2., 3.],
+            [4., 5.]
+        ])
+        ref_two_body = np.array([
```
ValentinS4t1qbit, line 68: Alternative: define as `np.zeros((2,2,2,2))` and set up the 3 only non-zero values explicitly.

tangelo/toolboxes/molecular_computation/molecule.py
```diff
@@ -65,6 +65,40 @@ def molecule_to_secondquantizedmolecule(mol, basis_set="sto-3g", frozen_orbitals
...
+        two_body_coefficients: Two-body coefficients (array of 2N*2N*2N*2N,
+            where N is the number of molecular orbitals).
+
+    Returns:
+        (array of floats, array of floats): One- and two-body integrals (arrays
+            of N*N and N*N*N*N elements, where N is the number of molecular
+            orbitals.
+    """
+    # Get the number of MOs = number of SOs / 2.
+    n_mos = int(one_body_coefficients.shape[0] / 2)
```
ValentinS4t1qbit, line 83: Use the `//` operator for division with integers, and then no need to cast to int.

ValentinS4t1qbit (approved):

ValentinS4t1qbit (approved):

PR #221, opened 2022-09-26
Relaxing iqcc_ilc_h4_cation test

AlexandreF-1qbit:
I noticed there are 2 possible solutions (-1.638999 and -1.638197) with the iqcc_ilc_h4_cation test. This is maybe an underlying consequence of the iQCC-ILC algorithm.


ValentinS4t1qbit (approved):
As long as we think the required accuracy of this test is still enough to claim that iQCC-ILC is working as expected, I have no issue with the modifications

PR #222, opened 2022-09-27
Qiskit runtime integration (IBM quantum only)

ValentinS4t1qbit:
Connection class for IBM Quantum, using qiskit runtime.
The initialization of the connection with IBM Cloud will occur later, when we have access to it.


tangelo/linq/qpu_connection/ibm_connection.py
```diff
@@ -0,0 +1,140 @@
...
+
+from tangelo.linq.translator import translate_qiskit
+from tangelo.linq.qpu_connection.qpu_connection import QpuConnection
+
+try:
+    from qiskit.providers.jobstatus import JobStatus
+    from qiskit_ibm_runtime import QiskitRuntimeService
+    is_qiskit_installed = True
+except ModuleNotFoundError:
+    is_qiskit_installed = False
```
AlexandreF-1qbit, line 16: Nice! I think I will take a similar approach for the `installation without pyscf` task. We can also change the encapsulation of `qemist_client` in the future!

tangelo/linq/tests/test_ibm_connection.py
```diff
@@ -0,0 +1,80 @@
...
+import time
+
+from tangelo.linq import Gate, Circuit
+from tangelo.linq.qpu_connection import IBMConnection
+
+circ1 = Circuit([Gate("H", 0), Gate("X", 1)])
+res_simulator_circ1 = {'01': 0.5, '11': 0.5}
+
+
+def assert_freq_dict_almost_equal(d1, d2, atol):
```
AlexandreF-1qbit, line 17: I saw this function many places. I think it would be better to put it in `tangelo.helpers` (a testing module?) and import it from there.

AlexandreF-1qbit (approved):

tangelo/linq/qpu_connection/ibm_connection.py
```diff
@@ -0,0 +1,195 @@
...
+
+    def job_submit(self, program, backend_name, circuits, operators=None, n_shots=10**4, runtime_options=None):
+        """ Submit job, return job ID.
+
+        Args:
+            program (str): name of available qiskit-runtime program (e.g sampler, estimator currently)
+            backend_name (str): name of a qiskit backend
+            circuits (Circuit | List[Circuit]): Tangelo circuit(s)
+            operators (QubitOperator | List[QubitOperator]) : Optional, qubit operators for computing expectation values
+            n_shots (int): Optional, number of shots to use on the target backend
```
AlexandreF-1qbit, line 56: Personally, I would let `n_shots` to be provided by the user. 10**4 is kind of in the low side. Also, is there a difference between `n_shots` using the sampler and the estimator (and specify it in the doc)? I would assume that the real number of shots for estimator is `n_shots * len(operators)`.

AlexandreF-1qbit (approved):
Only one thing about the `n_shots` specification. I am trusting you with whatever choice you decide (optional or not).

tangelo/linq/qpu_connection/ibm_connection.py
ValentinS4t1qbit: @AlexandreF-1qbit Yeah I think the number of shots should not be optional, as it directly relates to the amount of time / computation someone is going to be billed, and the accuracy of results. The user should have their eyes on it.

qiskit-runtime is still seeing a lot of changes, and the documentation and examples are a bit all over the place. I think Estimator may also do things behind the scenes (QWC, others), as each of these operators in `[operators]` can be a Hamiltonian.

I think the right course of action is requiring a number of shots from the user, and letting IBM Quantum do its thing.

JamesB-1qbit (approved):

tangelo/linq/qpu_connection/ibm_connection.py
AlexandreF-1qbit: All good with me!

PR #228, opened 2022-10-06
compute_rdms function

KrzysztofB-1qbit:
First draft of compute_rdms. Main changes:
- Added the function, can take classical shadow, Pauli expectation values, or raw data in the form {basis: histogram}
- Added tests for raw data and classical shadow inputs, including a raw data file
- Added n_spinorbitals, n_electrons, spin to FermionOperator class
- Added filter_bases helper function


tangelo/toolboxes/molecular_computation/rdms.py
```diff
@@ -41,6 +48,96 @@ def matricize_2rdm(two_rdm, n_orbitals):
...
+        qubit_term.compress()
+
+        # Loop to go through all qubit terms.
+        eigenvalue = 0.
+
+        if isinstance(exp_data, RandomizedClassicalShadow):
+            for qterm, coeff in qubit_term.terms.items():
+                if coeff.real != 0:
+                    # Change depending on if it is randomized or not.
+                    eigenvalue += exp_data.get_term_observable(qterm, coeff)
```
JamesB-1qbit, line 106: You should be able to write this code block as.
```python3
if issubclass(exp_data, ClassicalShadow):
     eigenvalue = exp_data.get_observable(qubit_term, coeff)
```

RandomizedClassicalShadow has the extra `k=10` option so you may want to have an extra Arg like `eval_args` as a dictionary and then unpack using `exp_data.get_observable(qubit_term, coeff, **eval_args)` where a user would give `eval_args={'k': 5}`

tangelo/toolboxes/molecular_computation/rdms.py
```diff
@@ -41,6 +48,96 @@ def matricize_2rdm(two_rdm, n_orbitals):
...
+
+        # Fermionic term with a prefactor of 1.0.
+        fermionic_term = FermionOperator(term, 1.0)
+
+        qubit_term = fermion_to_qubit_mapping(fermion_operator=fermionic_term,
+                                            n_spinorbitals=ferm_ham.n_spinorbitals,
+                                            n_electrons=ferm_ham.n_electrons,
+                                            mapping=mapping,
+                                            up_then_down=up_then_down)
+        qubit_term.compress()
```
JamesB-1qbit, line 97: 'scbk' requires spin so that should also be defined in ferm_ham

tangelo/toolboxes/molecular_computation/tests/test_rdms.py
```diff
@@ -14,39 +14,78 @@
...
 
 # For openfermion.load_operator function.
 pwd_this_test = os.path.dirname(os.path.abspath(__file__))
 
+ferm_op_of = load_operator("H2_ferm_op.data", data_directory=pwd_this_test + "/data", plain_text=True)
+ferm_op = FermionOperator()
+ferm_op.__dict__ = ferm_op_of.__dict__
+ferm_op.n_spinorbitals = 4
+ferm_op.n_electrons = 2
+ferm_op.spin = 0
```
JamesB-1qbit, line 35: What happens if you do
```python3
ferm_op = FermionOperator(n_spinorbitals=4, n_electrons=2, spin=0) + load_operator("H2_ferm_op.data", data_directory=pwd_this_test + "/data", plain_text=True)
```

tangelo/toolboxes/molecular_computation/rdms.py
KrzysztofB-1qbit: I did `isinstance` rather than `issubclass`, but other than that done.

tangelo/toolboxes/molecular_computation/rdms.py
KrzysztofB-1qbit: Done

tangelo/toolboxes/molecular_computation/tests/test_rdms.py
KrzysztofB-1qbit: I get 
`TypeError: Cannot add invalid type to <class 'tangelo.toolboxes.operators.operators.FermionOperator'>.`
Should that be fixed somehow?

JamesB-1qbit: Yes. If you look in the same file at QubitHamiltonian. You will need to add something like
```python3
    def __iadd__(self, other_hamiltonian):

        # Raise error if attributes are not the same across Hamiltonians. This
        # check is ignored if comparing to a QubitOperator or a bare
        # QubitHamiltonian.
        if self.mapping is not None and self.up_then_down is not None and \
                                other_hamiltonian.mapping is not None and \
                                other_hamiltonian.up_then_down is not None:

            if self.mapping.upper() != other_hamiltonian.mapping.upper():
                raise RuntimeError("Mapping must be the same for all QubitHamiltonians.")
            elif self.up_then_down != other_hamiltonian.up_then_down:
                raise RuntimeError("Spin ordering must be the same for all QubitHamiltonians.")

        return super(QubitOperator, self).__iadd__(other_hamiltonian)

    def __eq__(self, other_hamiltonian):

        # Additional checks for == operator. This check is ignored if comparing
        # to a QubitOperator or a bare QubitHamiltonian.
        if self.mapping is not None and self.up_then_down is not None and \
                                other_hamiltonian.mapping is not None and \
                                other_hamiltonian.up_then_down is not None:
            if (self.mapping.upper() != other_hamiltonian.mapping.upper()) or (self.up_then_down != other_hamiltonian.up_then_down):
                return False

        return super(QubitOperator, self).__eq__(other_hamiltonian)
```

tangelo/toolboxes/molecular_computation/rdms.py
```diff
@@ -41,6 +48,96 @@ def matricize_2rdm(two_rdm, n_orbitals):
...
+            for qterm, coeff in qubit_term.terms.items():
+                if coeff.real != 0:
+                    # Change depending on if it is randomized or not.
+                    eigenvalue += exp_data.get_term_observable(qterm, coeff)
+
+        if isinstance(exp_data, dict):
+            for qterm, coeff in qubit_term.terms.items():
+                if coeff.real != 0:
+
+                    try:
```
AlexandreF-1qbit, line 112: I would change this to something more dense:
```python
exp_val = 1. if not qterm else exp_vals.get(term, some_function(...))
```
Where
```python
def some_function(...):
    ps = pauli_of_to_string(qterm, n_qubits)
    hist = aggregate_histograms(*[Histogram(exp_data[basis]) for basis in filter_bases(ps, 
    exp_data.keys())])
    return hist.get_expectation_value(qterm, 1.)
```

tangelo/toolboxes/molecular_computation/rdms.py
```diff
@@ -41,6 +48,96 @@ def matricize_2rdm(two_rdm, n_orbitals):
...
+
+    # Construct spin-summed 1-RDM.
+    for i, j in it.product(range(onerdm.shape[0]), repeat=2):
+        onerdm_spinsum[i//2, j//2] += onerdm[i, j]
+
+    # Construct spin-summed 2-RDM.
+    for i, j, k, l in it.product(range(twordm.shape[0]), repeat=4):
+        twordm_spinsum[i//2, j//2, k//2, l//2] += twordm[i, j, k, l]
+
+    return onerdm, twordm, onerdm_spinsum, twordm_spinsum
```
AlexandreF-1qbit, line 138: Should we put flags to return spinsum RDMs or not?

tangelo/toolboxes/molecular_computation/tests/test_rdms.py
```diff
@@ -14,39 +14,78 @@
...
-            ]
 
+    def test_energy_from_rdms(self):
+        """Computes energy using known spin-summed 1RDM and 2RDM"""
         e_rdms = energy_from_rdms(ferm_op, rdm1, rdm2)
         self.assertAlmostEqual(e_rdms, -1.1372701, delta=1e-5)
 
+    def test_compute_rdms_from_raw_data(self):
+        """Load data and compute RDMs from frequency list"""
+        rdm1r, rdm2r, rdm1ssr, rdm2ssr = compute_rdms(ferm_op, exp_data, "scbk", False)
```
AlexandreF-1qbit, line 76: `up_then_down` is forced to `True` (last argument). Same for the other test.

tangelo/toolboxes/operators/operators.py
```diff
@@ -27,10 +27,17 @@
 
 
 class FermionOperator(openfermion.FermionOperator):
-    """Currently, this class is coming from openfermion. Can be later on be
+    """Currently, this class is coming from openfermion. Can later on be
     replaced by our own implementation.
     """
 
+    def __init__(self, term=None, coefficient=1., n_spinorbitals=None, n_electrons=None, spin=None):
```
AlexandreF-1qbit, line 34: We will have to take advantage of the new attributes!

tangelo/toolboxes/molecular_computation/rdms.py
```diff
@@ -41,6 +48,96 @@ def matricize_2rdm(two_rdm, n_orbitals):
     return rho
 
 
+def compute_rdms(ferm_ham, exp_data, mapping, up_then_down):
```
AlexandreF-1qbit, line 51: I am still thinking of how to support many possibilities to compute eigenvalues, with a nice interface. Another way of doing that in a more "modular" way would be to let the user provide a function to compute expectation value. We could let the user in charge of providing the right `exp_data` for the `get_exp_value_whatever...` function he/she would define. With that design, we would support direct evaluation, randomized CS, derandomized CS, ..., and so on.

Food for thought.

AlexandreF-1qbit (changes requested):
Some trivial things to fix. Some comments are more general feelings we could discuss in group.

tangelo/toolboxes/molecular_computation/tests/test_rdms.py
KrzysztofB-1qbit: Added these and other methods. Now FermionOperator can be mixed with openfermion version intuitively. Added tests for those and fixed one for SecondQuantizedMolecule that was failing

tangelo/toolboxes/molecular_computation/rdms.py
KrzysztofB-1qbit: Done

tangelo/toolboxes/molecular_computation/rdms.py
KrzysztofB-1qbit: Left as was, as per our discussion

tangelo/toolboxes/molecular_computation/rdms.py
```diff
@@ -41,6 +48,96 @@ def matricize_2rdm(two_rdm, n_orbitals):
...
+        complex array: spin-summed 1-RDM
+        complex array: spin-summed 2-RDM
+    """
+    onerdm = np.zeros((ferm_ham.n_spinorbitals,) * 2, dtype=complex)
+    twordm = np.zeros((ferm_ham.n_spinorbitals,) * 4, dtype=complex)
+    onerdm_spinsum = np.zeros((ferm_ham.n_spinorbitals//2,) * 2, dtype=complex)
+    twordm_spinsum = np.zeros((ferm_ham.n_spinorbitals//2,) * 4, dtype=complex)
+
+    # Check if the data dict is expectation values or raw frequency data in the {basis: hist} format
+    if isinstance(exp_data, dict) and isinstance(list(exp_data.values())[0], float):
```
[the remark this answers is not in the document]
KrzysztofB-1qbit: I defined 3 different arguments with default values of None. Conditionals enforce that only one of them can be defined. Rest of the code remains unchanged.

tangelo/toolboxes/molecular_computation/rdms.py
KrzysztofB-1qbit: Left for a separate PR, as per our discussion

tangelo/toolboxes/operators/operators.py
```diff
@@ -27,10 +28,54 @@
...
+        super(FermionOperator, self).__init__(term, coefficient)
+
+        self.n_spinorbitals = n_spinorbitals
+        self.n_electrons = n_electrons
+        self.spin = spin
+
+    def __iadd__(self, other):
+        # Raise error if attributes are not the same across Operators.
+        if (self.n_spinorbitals is not None and other.n_spinorbitals is not None)\
```
KrzysztofB-1qbit, line 43: I think this the right way of doing this check, as opposed to what QubitHamiltonian is doing

tangelo/toolboxes/operators/operators.py
```diff
@@ -27,10 +28,54 @@
...
+
+    def __eq__(self, other):
+        # Additional checks for == operator.
+        if isinstance(other, openfermion.FermionOperator):
+            return super(FermionOperator, self).__eq__(other)
+        elif (self.n_spinorbitals != other.n_spinorbitals) or (self.n_electrons != other.n_electrons) or (self.spin != other.spin):
+            return False
+
+        return super(FermionOperator, self).__eq__(other)
+
```
JamesB-1qbit, line 78: Do the operations `op -= newop`, `op-newop`, `newop-op`, `op*newop`, `newop*op` where `op` is our `FermionOperator` and `newop` is openfermion `FermionOperator`.

tangelo/toolboxes/molecular_computation/rdms.py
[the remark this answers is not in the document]
AlexandreF-1qbit: I am fine with that!

JamesB-1qbit (approved):

tangelo/toolboxes/operators/operators.py
```diff
@@ -20,17 +20,102 @@
 from collections import OrderedDict
 
 import numpy as np
+from openfermion.ops.operators.symbolic_operator import COEFFICIENT_TYPES
```
ValentinS4t1qbit, line 23: This is currently not used and Pycharm should be telling you that (code is gray)

tangelo/toolboxes/operators/tests/test_operators.py
```diff
@@ -73,6 +74,71 @@ def test_get_coeff(self):
         np.testing.assert_array_almost_equal(ref_one_body, one_body)
         np.testing.assert_array_almost_equal(ref_two_body, two_body)
 
+    def test_add_FermionOperator(self):
```
ValentinS4t1qbit, line 77: 1. The test class name already mention FermionOperator. You do not particularly need to repeat that into your test name, it is implicit.
2. Please add explicit tests for `__eq__` and `__neq__` in a new `test_eq` (using `==` and `!=`)

tangelo/toolboxes/operators/operators.py
```diff
@@ -20,17 +20,102 @@
...
+        return self.__isub__(other)
+
+    def __eq__(self, other):
+        # Additional checks for == operator.
+        if isinstance(other, of.FermionOperator):
+            return super(FermionOperator, self).__eq__(other)
+        elif (self.n_spinorbitals != other.n_spinorbitals) or (self.n_electrons != other.n_electrons) or (self.spin != other.spin):
+            return False
+
+        return super(FermionOperator, self).__eq__(other)
```
ValentinS4t1qbit, line 117: This statement does not make sense. Here's why:

For this statement to work, `self` must be a child class of `FermionOperator` (otherwise super does not work.)
But if that was the case, we fall into the branch at L112 and we return at L113. Therefore never hitting L117.

tangelo/toolboxes/operators/operators.py
```diff
@@ -20,17 +20,102 @@
...
+        return self.__iadd__(-1. * other)
+
+    def __sub__(self, other):
+        return self.__isub__(other)
+
+    def __rsub__(self, other):
+        return self.__isub__(other)
+
+    def __eq__(self, other):
+        # Additional checks for == operator.
```
ValentinS4t1qbit, line 111: Right now:

- Because `isinstance` will return `True` for any class that inherits from `of.FermionOperator` I make the hypothesis that when you call `f_op1 == f_op2` it actually never enters the second branch and will return `True` as long as terms match, even if there's a disagreement in spin etc. This is some of the object-oriented programming you want to get better at: because `op.FO` (for short) is a parent class of `FO`, it means any FO instance (object)is also an instance of `op.FO`. So `isinstance` returns `True` for both.

Knowing that we can be be crafty. I think this code block can become your entire function

```
if isinstance(other,  FermionOperator) : # Only works if it's a Tangelo FO, or any future child classes of it.
   a = super(FermionOperator, self).__eq__(super(FermionOperator, other))
   b = (self.n_spinorbitals, self.n_electrons, self.spin) == (other.n_spinorbitals, other.n_electrons, other.spin)
   return a and b
elif isinstance(other,  of.FermionOperator)  # Can't be a Tangelo FO, otherwise we got into the 1st branch
   return super(FermionOperator, self).__eq__(other)
else: # Should raise an error
    Raise TypeError(f"Cannot compare objects of type {self.__class__} and {other.__class__}")
```

ValentinS4t1qbit (changes requested):
Tough PR if you're not familiar with object-oriented programming but excellent learning experience: don't give up !

Do not hesitate to reach out if you have any question, it is easier to explain orally.

tangelo/toolboxes/operators/operators.py
```diff
@@ -20,17 +20,102 @@
...
+            return super(FermionOperator, self).__imul__(other)
+        elif isinstance(other, of.FermionOperator):
+            if self.n_spinorbitals is not None or self.n_electrons is not None or self.spin is not None:
+                raise RuntimeError("openfermion FermionOperator did not define a necessary attribute")
+            else:
+                f_op = FermionOperator()
+                f_op.terms = other.terms.copy()
+                return super(FermionOperator, self).__imul__(f_op)
+        else:
+            return super(FermionOperator, self).__imul__(other)
```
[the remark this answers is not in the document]
JamesB-1qbit: I think this relies on the type check of openfermion where other could be a coefficient.

tangelo/toolboxes/operators/operators.py
```diff
@@ -21,16 +21,89 @@
...
+                return super(FermionOperator, self).__imul__(f_op)
+
+        else:
+            return super(FermionOperator, self).__imul__(other)
+
+    def __mul__(self, other):
+        return self.__imul__(other)
+
+    def __rmul__(self, other):
+        return self.__imul__(other)
```
JamesB-1qbit, line 61: I don't think you need to redefine this. From openfermion `SymbolicOperator`

```python3
    def __rmul__(self, multiplier):
        """
        Return multiplier * self for a scalar.

        We only define __rmul__ for scalars because the left multiply
        exist for  SymbolicOperator and left multiply
        is also queried as the default behavior.

        Args:
          multiplier: A scalar to multiply by.

        Returns:
          product: A new instance of SymbolicOperator.

        Raises:
          TypeError: Object of invalid type cannot multiply SymbolicOperator.
        """
        if not isinstance(multiplier, COEFFICIENT_TYPES):
            raise TypeError('Object of invalid type cannot multiply with ' +
                            type(self) + '.')
        return self * multiplier
```

ValentinS4t1qbit (changes requested):
There are some conversations we may need a talk to solve as I am not clear on why things were implemented the way they currently are.

tangelo/toolboxes/operators/operators.py
KrzysztofB-1qbit: good point

tangelo/toolboxes/molecular_computation/rdms.py
```diff
@@ -41,6 +48,108 @@ def matricize_2rdm(two_rdm, n_orbitals):
...
+    """
+    Return the 1- and 2-RDM and their spin-summed versions
+    using a Molecule object and frequency data either in the form of a
+    classical shadow or a dictionary of frequency histograms.
+
+    Args:
+        ferm_ham (FermionicOperator): Fermionic operator with n_spinorbitals, n_electrons, and spin defined
+        mapping (str): Qubit mapping
+        up_then_down (bool): Spin ordering for the mapping
+        exp_vals (dict): Dictionary of Pauli word expectation values
```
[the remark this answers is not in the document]
KrzysztofB-1qbit: Are you sure?

ValentinS4t1qbit (approved):
Great job ! That was a lot of work, and I hope you've learnt a bunch in the process.

PR #229, opened 2022-10-12
add __repr__ method to Gate

JamesB-1qbit:
Some of our functions return a list of gates such as measurement_basis_gates. Without a repr method printing this return results in the output
```
>>> from tangelo.linq import Gate
>>> from tangelo.linq.helpers.circuits import measurement_basis_gates
>>> m = measurement_basis_gates(((0, "Y"), (1, "X")))
>>> print(m)
[{}, {}]
>>> m
[{}, {}]
```
With repr, the following output occurs.
```
>>> from tangelo.linq.helpers.circuits import measurement_basis_gates
>>> m = measurement_basis_gates(((0, "Y"), (1, "X")))
>>> print(m)
[Gate(name='RX', target=[0], parameter=1.5707963267948966), Gate(name='RY', target=[1], parameter=-1.5707963267948966)]
>>> m
[Gate(name='RX', target=[0], parameter=1.5707963267948966), Gate(name='RY', target=[1], parameter=-1.5707963267948966)]
```
Also `eval(gate.__repr__()) == gate`.

### DIFF
--- a/tangelo/linq/gate.py
+++ b/tangelo/linq/gate.py
@@ -120,6 +120,22 @@ class Gate(dict):
 
         return mystr
 
+    def __repr__(self):
+        """Represent gate such that eval(self.__repr__) returns a proper instance of Gate."""
+
+        mystr = f"Gate(name='{self.name}'"
+        for attr in ["target", "control"]:
+            if self.__getattribute__(attr) or isinstance(self.__getattribute__(attr), int):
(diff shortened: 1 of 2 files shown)


ValentinS4t1qbit (approved):
Good stuff, thank you for improving that ! I see no reason to not turn this into a PR and merge it if tests pass.

PR #230, opened 2022-10-13
VQESolver cleanup

AlexandreF-1qbit:
Highlights:
- Use of the `dict.update` method to update `backend_options`.
- Grouped ansatz building with the `eval` command.
- Added `reference_state` to ansatz arguments: all ansatze have more or less the same interface.
- The excited_states notebook was applying the k=2 option to UCCSD, raising an error in the new implementation (before the ansatz_options for UCCSD were ignored). I removed that part.

I left the "no-pyscf" option for a future PR (the one decoupling pyscf installation with tangelo).


JamesB-1qbit (approved):

tangelo/algorithms/variational/vqe_solver.py
```diff
@@ -213,43 +208,20 @@ def build(self):
...
 
         # Building with a qubit Hamiltonian.
-        elif self.ansatz in [BuiltInAnsatze.HEA, BuiltInAnsatze.VSQS]:
-            if self.ansatz == BuiltInAnsatze.HEA:
-                self.ansatz = HEA(self.molecule, self.qubit_mapping, self.up_then_down, **self.ansatz_options)
-            elif self.ansatz == BuiltInAnsatze.VSQS:
-                self.ansatz = VSQS(self.molecule, self.qubit_mapping, self.up_then_down, **self.ansatz_options)
-        elif not isinstance(self.ansatz, Ansatz):
+        elif self.ansatz in {BuiltInAnsatze.HEA, BuiltInAnsatze.VSQS}:
+            self.ansatz = eval(f"agen.{self.ansatz.name}(self.molecule, self.qubit_mapping, self.up_then_down, **self.ansatz_options)")
```
ValentinS4t1qbit, line 223: Pretty ugly way to do it.

I'd rather you have a dictionary `d` that maps the enum `BuiltInAnsatze` values to the name of the class you are instantiating. Then the class name is given by `d[BuiltInAnsatze.HEA]` and you can instantiate with `d[BuiltInAnsatze.HEA](...)`

If this dictionary is mostly trivially defined using the `name` attribute, so be it.

tangelo/algorithms/variational/vqe_solver.py
```diff
@@ -213,43 +208,20 @@ def build(self):
...
-                elif self.ansatz == BuiltInAnsatze.UCCGD:
-                    self.ansatz = UCCGD(self.molecule, self.qubit_mapping, self.up_then_down, **self.ansatz_options)
-                    self.ansatz.default_reference_state = "HF" if self.ref_state is None else "zero"
-                elif self.ansatz == BuiltInAnsatze.ILC:
-                    self.ansatz = ILC(self.molecule, self.qubit_mapping, self.up_then_down, **self.ansatz_options)
+                if self.ansatz in {BuiltInAnsatze.UCC1, BuiltInAnsatze.UCC3}:
+                    n_var_params = 1 if self.ansatz == BuiltInAnsatze.UCC1 else 3
+                    self.ansatz = agen.RUCC(n_var_params)
+                elif self.ansatz in self.builtin_ansatze:
+                    self.ansatz = eval(f"agen.{self.ansatz.name}(self.molecule, self.qubit_mapping, self.up_then_down, **self.ansatz_options)")
```
ValentinS4t1qbit, line 215: see comment below

ValentinS4t1qbit (changes requested):
Good, just 1-2 things.

tangelo/algorithms/variational/vqe_solver.py
AlexandreF-1qbit: The enum class itself is kind of a dictionary. Creating another one would be redundant according to me. Now we don't deal with the `int` in the `BuiltInAnsatze`, and we could change those values to the objects.
```python
class BuiltInAnsatze(Enum):
    UCCSD = agen.UCCSD
    UCC1 = agen.RUCC
    UCC3 = agen.RUCC
    HEA = agen.HEA
...
```
And then we could do (if `self.ansatz` is a `BuiltInAnsatze` -> `self.ansatz(self.molecule, ...)`. This was the other option I considered.

ValentinS4t1qbit (approved):
Thank you for helping with this refactoring and cleanup !

PR #232, opened 2022-10-18
Symmetry post-selection

KrzysztofB-1qbit:
Adds the symmetry post-selection functionality
- Adds ancilla_symmetry_circuit function to append the symmetry verification circuit to a given circuit
- Adds post_select function to pst select on a dictionary of ancilla qubits
- Adds a convenience strip_post_selection function to remove the ancilla and aggregate the data as if no post-selection was used
- Adds tests for the functions

### DIFF
--- a/tangelo/toolboxes/post_processing/post_selection.py
+++ b/tangelo/toolboxes/post_processing/post_selection.py
@@ -19,6 +19,7 @@ import warnings
 from tangelo.linq import Circuit, Gate
 from tangelo.linq.helpers import measurement_basis_gates, pauli_string_to_of
 from tangelo.toolboxes.operators import QubitOperator, count_qubits
+from tangelo.toolboxes.post_processing import Histogram
 
 
 def ancilla_symmetry_circuit(circuit, sym_op):
@@ -60,12 +61,12 @@ def ancilla_symmetry_circuit(circuit, sym_op):
     return circuit_new
 
 
-def post_select(hist, expected_outcomes):
+def post_select(freqs, expected_outcomes):
     """Apply post selection to frequency data based on
     a dictionary of expected outcomes on ancilla qubits.
 
     Args:
-        hist (dict): A dictionary of {bitstring: frequency} pairs
+        freqs (dict): A dictionary of {bitstring: frequency} pairs
         expected_outcomes (dict): Desired outcomes on certain qubit indices
             and their expected state. For example, {0: "1", 1: "0"} would
             filter results based on the first qubit with the |1> state and
@@ -75,41 +76,21 @@ def post_select(hist, expected_outcomes):
         dict: A dictionary of post-selected, renormalized frequencies
             and bitstrings with removed ancilla qubits
     """
-    def f_post_select(bitstring):
-        for qubit_i, expected_bit in expected_outcomes.items():
-            if bitstring[qubit_i] != expected_bit:
-                return False
-        return True
-
-    counts_new = {}
-
-    for string, freq in hist.items():
-        if f_post_select(string):
-            new_str = "".join([s for i, s in enumerate(string) if i not in expected_outcomes.keys()])
-            counts_new[new_str] = freq
-
-    factor = sum(counts_new.values())
-    counts_new = {key: value/factor for key, value in counts_new.items()}
-    return counts_new
+    hist = Histogram(freqs, n_shots=0)
+    hist.post_select(expected_outcomes)
+    return hist.frequencies
 
 
-def strip_post_selection(freqs, qubits):
+def strip_post_selection(freqs, *qubits):
     """Convenience function to remove the symmetry ancilla qubit
     and aggregate data to recreate results without post-selection.
 
     Args:
         freqs (dict): A dictionary of {bitstring: frequency} as returned from a quantum device
-        qubits (int or list): An ancilla qubit index, or list of indices, to be removed from the bitstrings
+        qubits (variable number of int): The ancilla qubit indices to be removed from the bitstrings
 
     Returns:
         dict: A frequency dictionary with the qubits stripped and data aggregated"""
-    if isinstance(qubits, int):
-        qubits = [qubits]
-
-    counts_new = {}
-
-    for string, counts in freqs.items():
-        new_str = "".join([s for i, s in enumerate(string) if i not in qubits])
-        counts_new[new_str] = counts_new.get(new_str, 0) + counts
-
-    return counts_new
+    hist = Histogram(freqs, n_shots=0)
+    hist.remove_qubit_indices(*qubits)
+    return hist.frequencies
--- a/tangelo/toolboxes/post_processing/post_selection.py
+++ b/tangelo/toolboxes/post_processing/post_selection.py
@@ -12,24 +12,48 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-"""This module provides functions to create symmetry verification and postselection circuits."""
+"""This module provides functions to create symmetry verification and post-selection circuits."""
+
+import warnings
+
 from tangelo.linq import Circuit, Gate
 from tangelo.linq.helpers import measurement_basis_gates, pauli_string_to_of
+from tangelo.toolboxes.operators import QubitOperator, count_qubits
 
 
 def ancilla_symmetry_circuit(circuit, sym_op):
-    """Appends to a circuit a symmetry operator and ancilla for symmetry verification.
+    """Append a symmetry operator circuit to an input circuit using an extra ancilla qubit,
+    for the purpose of symmetry verification and post-selection.
     For more details see arXiv:1807.10050.
 
     Args:
         circuit (Circuit): A quantum circuit to be equipped with symmetry verification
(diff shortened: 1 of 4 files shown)


tangelo/toolboxes/post_processing/post_selection.py
```diff
@@ -0,0 +1,94 @@
...
+            and entanglement with an ancilla qubit"""
+    n_qubits = circuit.width
+    basis_gates = measurement_basis_gates(pauli_string_to_of(sym_op))
+    basis_circ = Circuit(basis_gates)
+    parity_gates = [Gate("CNOT", n_qubits, i) for i in range(n_qubits)]
+    circuit_new = circuit + basis_circ + Circuit(parity_gates) + basis_circ.inverse()
+    return circuit_new
+
+
+def post_select(hist, expected_outcomes):
```
AlexandreF-1qbit, line 39: Is there a reason we want a post_select function taking frequencies and ignoring the number of shots? If not, we could only reuse the `Histogram.post_select` function and delete this function.

AlexandreF-1qbit (changes requested):
I feel some functions can be rewritten with the Histogram class and an arbitrary number of shots (I think specifying `n_shots=1` would work).

tangelo/toolboxes/post_processing/post_selection.py
```diff
@@ -0,0 +1,94 @@
...
+from tangelo.linq.helpers import measurement_basis_gates, pauli_string_to_of
+
+
+def ancilla_symmetry_circuit(circuit, sym_op):
+    """Appends to a circuit a symmetry operator and ancilla for symmetry verification.
+    For more details see arXiv:1807.10050.
+
+    Args:
+        circuit (Circuit): A quantum circuit to be equipped with symmetry verification
+        sym_op (string): The symmetry operator to be verified, in the form of a Pauli string
```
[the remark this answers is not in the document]
ValentinS4t1qbit: Not fond of the string being what we expect from the user. As Alexandre said, we tend to manipulate `QubitOperator` objects, or openfermion-style (tuple) representations (which causes you to use `pauli_string_to_of`).

I'm not sure why we'd support the string case, I'd need to be convinced. I think for consistency with the rest of the package, we should not.

tangelo/toolboxes/post_processing/post_selection.py
```diff
@@ -0,0 +1,94 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""This module provides functions to create symmetry verification and postselection circuits."""
+from tangelo.linq import Circuit, Gate
+from tangelo.linq.helpers import measurement_basis_gates, pauli_string_to_of
+
+
+def ancilla_symmetry_circuit(circuit, sym_op):
+    """Appends to a circuit a symmetry operator and ancilla for symmetry verification.
```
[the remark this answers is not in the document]
ValentinS4t1qbit: How about we say

`"""Append a symmetry operator circuit using an extra ancilla qubit, for the purpose of symmetry verification, to the input circuit."""`

tangelo/toolboxes/post_processing/post_selection.py
```diff
@@ -0,0 +1,94 @@
...
+    n_qubits = circuit.width
+    basis_gates = measurement_basis_gates(pauli_string_to_of(sym_op))
+    basis_circ = Circuit(basis_gates)
+    parity_gates = [Gate("CNOT", n_qubits, i) for i in range(n_qubits)]
+    circuit_new = circuit + basis_circ + Circuit(parity_gates) + basis_circ.inverse()
+    return circuit_new
+
+
+def post_select(hist, expected_outcomes):
+    """Applies post selection to frequency data based on
```
ValentinS4t1qbit, line 40: The custom is not to conjugate the verbs in docstrings and comments, when you start with a verb like that.
`Apply post-selection to....`

tangelo/toolboxes/post_processing/post_selection.py
ValentinS4t1qbit: Whatever we do, let's avoid defining redundant code in several places and maybe try to write one functionality using the other, or as a special case or the other.

But first and foremost: let's clarify what this new functionality brings, that we did not have before.

ValentinS4t1qbit (changes requested):
We need to figure out if some of these new functions implement a useful feature. If they do, how can be reuse similar existing code to avoid redundancy (e.g can we implement a functionality by reusing code from the other, or expressing it as a special case of the other).

Note: there are no tests for `ancilla_symmetry_circuit`. Are there any corner cases or tricky situations we can think of, that would warrant writing a test ?

If not, at the very least you should have tried an input circuit, shoved it into that function and drawn the output to see that it was what you expected.

tangelo/toolboxes/post_processing/post_selection.py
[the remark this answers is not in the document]
KrzysztofB-1qbit: I was thinking it might be more intuitive for the user to use Pauli strings. This is more like a human readable flag which would normally be used "by hand", where the user takes a circuit and decides which symmetry to input. It's much easier for a human to say "XX" than [(0, 'X'), (1, 'X')] or construct a whole QubitOperator. I can add functionality to work with other types just in case someone needs to do that though. 

Also, could add some consistency checks to make sure the number of qubits matches the input circuit.

tangelo/toolboxes/post_processing/post_selection.py
```diff
@@ -0,0 +1,96 @@
...
+
+    # Check if the operator size matches the circuit width
+    if n_qubits < op_len:
+        raise RuntimeError("The size of the symmetry operator is bigger than the circuit width.")
+    elif n_qubits > op_len:
+        warnings.warn("The size of the symmetry operator is smaller than the circuit width. Remaining qubits will be measured in the Z-basis.")
+
+    if isinstance(sym_op, str):
+        basis_gates = measurement_basis_gates(pauli_string_to_of(sym_op))
+    elif isinstance(sym_op, list or tuple):
```
ValentinS4t1qbit, line 53: I've never seen this for testing several types, and I don't know what it actually does.
I recommend `(list, tuple)` as your second argument, as it works for sure and is commonly recommended: https://stackoverflow.com/questions/33311258/python-check-if-variable-isinstance-of-any-type-in-list

tangelo/toolboxes/post_processing/post_selection.py
```diff
@@ -0,0 +1,96 @@
...
+            and entanglement with an ancilla qubit, which is added as the last qubit.
+            Increases the input circuit width by 1."""
+    if isinstance(sym_op, QubitOperator):
+        op_len = count_qubits(sym_op)
+    else:
+        op_len = len(sym_op)
+    n_qubits = circuit.width
+
+    # Check if the operator size matches the circuit width
+    if n_qubits < op_len:
```
ValentinS4t1qbit, line 46: In both your error or warning messages, I think we can talk about number of qubits instead of width.

tangelo/toolboxes/post_processing/post_selection.py
```diff
@@ -0,0 +1,96 @@
...
+        raise RuntimeError("The size of the symmetry operator is bigger than the circuit width.")
+    elif n_qubits > op_len:
+        warnings.warn("The size of the symmetry operator is smaller than the circuit width. Remaining qubits will be measured in the Z-basis.")
+
+    if isinstance(sym_op, str):
```
ValentinS4t1qbit, line 51: You have no `else` here. I don't think it's something you've done in this function: if the `symptoms_op` is none of these 3 things then we should return an error. That should cover our backs if the user pushes in something of the wrong type.

ValentinS4t1qbit (changes requested):
Some very minor things that should be straightforward. Almost there !

tangelo/toolboxes/post_processing/post_selection.py
```diff
@@ -0,0 +1,96 @@
...
+
+    Args:
+        circuit (Circuit): A quantum circuit to be equipped with symmetry verification
+        sym_op (string): The symmetry operator to be verified.
+            Can be a Pauli string, OpenFermion style list or a QubitOperator
+
+    Returns:
+        Circuit: The input circuit appended with the proper basis rotation
+            and entanglement with an ancilla qubit, which is added as the last qubit.
+            Increases the input circuit width by 1."""
```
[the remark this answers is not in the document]
KrzysztofB-1qbit: ok

tangelo/toolboxes/post_processing/post_selection.py
KrzysztofB-1qbit: ok

tangelo/toolboxes/post_processing/post_selection.py
KrzysztofB-1qbit: changed

tangelo/toolboxes/post_processing/post_selection.py
KrzysztofB-1qbit: Added error message

ValentinS4t1qbit (approved):
Thank you Krzys :)

AlexandreF-1qbit (approved):

PR #234, opened 2022-10-19
trotter order and  active electrons for QCC/iQCC

JamesB-1qbit:
Various bug fixes for iQCC.

- Trotter ordering not captured properly
- Used n_electrons instead of n_active_electrons.


ValentinS4t1qbit (approved):
Thank you James !

PR #235, opened 2022-10-20
Add circuit.depth() in get_resources methods

KrzysztofB-1qbit:
Added calls to circuit.depth() in get_resources method of vqe_solver, sa_vqe_solver and quantum_imaginary_time. No other instances were identified

### DIFF
--- a/tangelo/algorithms/projective/quantum_imaginary_time.py
+++ b/tangelo/algorithms/projective/quantum_imaginary_time.py
@@ -311,6 +311,5 @@ class QITESolver:
         resources["pool_size"] = len(self.pool_operators)
         resources["circuit_width"] = self.final_circuit.width
         resources["circuit_depth"] = self.final_circuit.depth()
-        resources["circuit_gates"] = self.final_circuit.size
         resources["circuit_2qubit_gates"] = self.final_circuit.counts_n_qubit.get(2, 0)
         return resources
--- a/tangelo/algorithms/projective/quantum_imaginary_time.py
+++ b/tangelo/algorithms/projective/quantum_imaginary_time.py
@@ -312,5 +312,5 @@ class QITESolver:
         resources["circuit_width"] = self.final_circuit.width
         resources["circuit_depth"] = self.final_circuit.depth()
         resources["circuit_gates"] = self.final_circuit.size
-        resources["circuit_2qubit_gates"] = self.final_circuit.counts.get("CNOT", 0)
+        resources["circuit_2qubit_gates"] = self.final_circuit.counts_n_qubit.get(2, 0)
         return resources
--- a/tangelo/algorithms/projective/quantum_imaginary_time.py
+++ b/tangelo/algorithms/projective/quantum_imaginary_time.py
@@ -310,6 +310,7 @@ class QITESolver:
         resources["qubit_hamiltonian_terms"] = len(self.qubit_hamiltonian.terms)
         resources["pool_size"] = len(self.pool_operators)
         resources["circuit_width"] = self.final_circuit.width
+        resources["circuit_depth"] = self.final_circuit.depth()
         resources["circuit_gates"] = self.final_circuit.size
         resources["circuit_2qubit_gates"] = self.final_circuit.counts.get("CNOT", 0)
         return resources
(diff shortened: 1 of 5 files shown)


AlexandreF-1qbit (approved):

tangelo/algorithms/projective/quantum_imaginary_time.py
```diff
@@ -310,6 +310,7 @@ def get_resources(self):
         resources["qubit_hamiltonian_terms"] = len(self.qubit_hamiltonian.terms)
         resources["pool_size"] = len(self.pool_operators)
         resources["circuit_width"] = self.final_circuit.width
+        resources["circuit_depth"] = self.final_circuit.depth()
         resources["circuit_gates"] = self.final_circuit.size
         resources["circuit_2qubit_gates"] = self.final_circuit.counts.get("CNOT", 0)
```
JamesB-1qbit, line 315: Since you are touching this function anyway. It has been on our todos to make the 2qubit gates more comprehensive. You could use https://github.com/goodchemistryco/Tangelo/blob/c88a97ef4e16a5e8d7d5c718eb7ae00d195c166d/tangelo/linq/gate.py#L25.

Although many of these could be in theory more than 2 qubit gates. This may be better done as a method in `Circuit`. And probably even better in Circuit.add_gate(). Something like adding the line
```python3
def Circuit:
     def __init__(self, gates: List[Gate] = None, n_qubits=None, name="no_name"):
          ...
          self._n_qubit_gate_counts = dict()
      ...
     def add_gate(self, g):
          ...
          n_qubit = len(g.control) + len(g.target)
          self._n_qubit_gate_counts[n_qubit] = self._n_qubit_gate_counts.get(n_qubit, 0) + 1
```

tangelo/algorithms/projective/quantum_imaginary_time.py
```diff
@@ -310,6 +310,7 @@ def get_resources(self):
         resources["qubit_hamiltonian_terms"] = len(self.qubit_hamiltonian.terms)
         resources["pool_size"] = len(self.pool_operators)
         resources["circuit_width"] = self.final_circuit.width
+        resources["circuit_depth"] = self.final_circuit.depth()
```
ValentinS4t1qbit, line 313: L313 is meant to replace L314. The number of gates ("size") of the circuit is not meaningful, the depth is.
You can remove L314 and do the same in the 2 other solvers

tangelo/algorithms/projective/quantum_imaginary_time.py
ValentinS4t1qbit: Counts for individual gates are already tracked in `self._gate_counts`.

I think I like James's idea: since Tangelo allows the user to build a circuit with any kind of gate (it's not a problem until it's converted to a particular format which may or may not support these gates), maybe it's better not to restrict ourselves to a list of predefined ones. Counting the number of qubits involved and then incrementing the corresponding count seems good and cheap.

tangelo/linq/circuit.py
```diff
@@ -173,6 +179,10 @@ def check_index_valid(index):
         # Keep track of the total gate count
         self._gate_counts[gate.name] = self._gate_counts.get(gate.name, 0) + 1
 
+        # Keep track of n_qubit gate count
```
ValentinS4t1qbit, line 182: Usually I'm pretty concise but I think it would be clearer for the reader if the comment was something like

`# Keep track of gate counts separately for 1-qubit gates, 2-qubit gates, and so on`

tangelo/linq/circuit.py
```diff
@@ -126,6 +127,11 @@ def counts(self):
         """Return the counts for all gate types included in the circuit."""
         return self._gate_counts
 
+    @property
+    def counts_n_qubit(self):
+        """Return the counts for all gate types included in the circuit."""
```
ValentinS4t1qbit, line 132: You just copy-pasted the docstring from L127. Let's turn this into something better:

`"""Return the number of 1-qubit gates, 2-qubit gates, and so on."""`

ValentinS4t1qbit (changes requested):
Good start. Just a few changes to make, and it should be good to go.

ValentinS4t1qbit (approved):
Looks good to me, thank you !

tangelo/algorithms/projective/quantum_imaginary_time.py
KrzysztofB-1qbit: Implemented as above for all n qubit gates, `get_resources` displays only the 2 qubit gate count

PR #238, opened 2022-10-22
Compute variance and standard error

jjgoings:
Adds functionality to compute the variance and standard error of an expectation value.

- Adds `get_variance` function to `Simulator` to compute the variance of expectation value
- Adds `get_standard_error` function to `Simulator` to compute the standard error from the mean (shot-dependent)
- Adds simple analytical test to check variance is computed correctly

### DIFF
--- a/tangelo/linq/simulator_base.py
+++ b/tangelo/linq/simulator_base.py
@@ -428,21 +428,17 @@ class SimulatorBase(abc.ABC):
                 variance += 0.
                 continue
 
-            if not self.n_shots:
-                # Without shots, variance using statevector is identically zero
-                variance += 0.
-
+            # If n_shots has been specified, then draw that amount of samples from the distribution
+            # and compute variance with empirical frequencies instead. Otherwise, use the exact frequencies
+            basis_circuit = Circuit(measurement_basis_gates(term), n_qubits=state_prep_circuit.width)
+            if basis_circuit.size > 0:
+                frequencies, _ = self.simulate(basis_circuit, initial_statevector=prepared_state)
             else:
-                # Run simulation with statevector but compute expectation value with samples directly drawn from it
-                basis_circuit = Circuit(measurement_basis_gates(term), n_qubits=state_prep_circuit.width)
-                if basis_circuit.size > 0:
-                    frequencies, _ = self.simulate(basis_circuit, initial_statevector=prepared_state)
-                else:
-                    frequencies = prepared_frequencies
-                variance_term = self.get_variance_from_frequencies_oneterm(term, frequencies)
-                # Assumes no correlation between terms
-                # https://en.wikipedia.org/wiki/Propagation_of_uncertainty#Example_formulae
-                variance += coef * coef * variance_term
+                frequencies = prepared_frequencies
+            variance_term = self.get_variance_from_frequencies_oneterm(term, frequencies)
+            # Assumes no correlation between terms
+            # https://en.wikipedia.org/wiki/Propagation_of_uncertainty#Example_formulae
+            variance += coef * coef * variance_term
 
         return variance
 
--- a/tangelo/linq/simulator_base.py
+++ b/tangelo/linq/simulator_base.py
@@ -77,6 +77,7 @@ def get_expectation_value_from_frequencies_oneterm(term, frequencies):
 
     return expectation_term
 
+
 def get_variance_from_frequencies_oneterm(term, frequencies):
     """Return the variance of the expectation value of a single-term qubit-operator, given
     the result of a state-preparation.
@@ -443,7 +444,7 @@ class SimulatorBase(abc.ABC):
                 # https://en.wikipedia.org/wiki/Propagation_of_uncertainty#Example_formulae
                 variance += coef * coef * variance_term
 
-        return variance 
+        return variance
 
     def _get_expectation_value_from_frequencies(self, qubit_operator, state_prep_circuit, initial_statevector=None):
         r"""Take as input a qubit operator H and a state preparation returning a
--- a/tangelo/linq/simulator_base.py
+++ b/tangelo/linq/simulator_base.py
@@ -280,6 +280,8 @@ class SimulatorBase(abc.ABC):
             qubit_operator(openfermion-style QubitOperator class): a qubit
                 operator.
             state_prep_circuit: an abstract circuit used for state preparation.
+            initial_statevector(list/array) : A valid statevector in the format
+                supported by the target backend.
 
         Returns:
             complex: The variance of this operator with regards to the
@@ -300,11 +302,7 @@ class SimulatorBase(abc.ABC):
 
         # If the underlying operator is hermitian, expectation value is real and can be computed right away
         if are_coefficients_real:
-            if self._noise_model or not self.statevector_available \
-                    or state_prep_circuit.is_mixed_state or state_prep_circuit.size == 0:
-                return self._get_variance_from_frequencies(qubit_operator, state_prep_circuit, initial_statevector=initial_statevector)
-            elif self.statevector_available:
-                return self._get_variance_from_statevector(qubit_operator, state_prep_circuit, initial_statevector=initial_statevector)
+            return self._get_variance_from_frequencies(qubit_operator, state_prep_circuit, initial_statevector=initial_statevector)
 
         # Else, separate the operator into 2 hermitian operators, use linearity and call this function twice
         else:
@@ -333,17 +331,15 @@ class SimulatorBase(abc.ABC):
             qubit_operator(openfermion-style QubitOperator class): a qubit
                 operator.
             state_prep_circuit: an abstract circuit used for state preparation.
+            initial_statevector(list/array) : A valid statevector in the format
+                supported by the target backend.
 
         Returns:
             complex: The standard error of this operator with regards to the
                 state preparation.
         """
         variance = self.get_variance(qubit_operator, state_prep_circuit, initial_statevector)
-
-        if not self.n_shots:
-            return 0.
-        else:
-            return np.sqrt(variance/self.n_shots)
+        return np.sqrt(variance/self.n_shots) if self.n_shots else 0.
 
     def _get_expectation_value_from_statevector(self, qubit_operator, state_prep_circuit, initial_statevector=None):
         r"""Take as input a qubit operator H and a state preparation returning a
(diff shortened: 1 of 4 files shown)


tangelo/linq/simulator_base.py
```diff
@@ -284,6 +398,54 @@ def _get_expectation_value_from_statevector(self, qubit_operator, state_prep_cir
...
+        # Otherwise, use generic statevector expectation value
+        for term, coef in qubit_operator.terms.items():
+
+            if len(term) > n_qubits:  # Cannot have a qubit index beyond circuit size
+                raise ValueError(f"Size of operator {qubit_operator} beyond circuit width ({n_qubits} qubits)")
+            elif not term:  # Empty term: no simulation needed
+                variance += 0.
+                continue
+
+            if not self.n_shots:
```
jjgoings, line 431: I don't think this is true. It's fine for use in computing standard error, but variance can still be non-zero with statevector. I'm not sure the best way to compute variance here in the exact case. Do Var(H) = <H^2> - <H>^2 directly?

AlexandreF-1qbit: Can something like https://github.com/goodchemistryco/Tangelo/blob/7a32ef2e72f24fd393b010e0a8500da4afae6188/tangelo/linq/simulator_base.py#L347
could be used in that situation? I mean just converting the statevector to ideal frequencies, and then resuse `_get_variance_from_frequencies` to get the ideal variance.

AlexandreF-1qbit (commented):
I just left one comment in the thread. Beside this, I am all good with the code (I will let you make changes if necessary and mark it as approved soon after that)!

tangelo/linq/simulator_base.py
```diff
@@ -231,6 +265,86 @@ def get_expectation_value(self, qubit_operator, state_prep_circuit, initial_stat
...
+        draw samples if required. In the case of a noisy simulator, or if the
+        statevector is not available on the target backend, individual shots
+        must be run and the workflow is akin to what we would expect from an
+        actual QPU.
+
+        Args:
+            qubit_operator(openfermion-style QubitOperator class): a qubit
+                operator.
+            state_prep_circuit: an abstract circuit used for state preparation.
+
```
ValentinS4t1qbit, line 283: I think we need to document `initial_statevector` as well

tangelo/linq/simulator_base.py
```diff
@@ -231,6 +265,86 @@ def get_expectation_value(self, qubit_operator, state_prep_circuit, initial_stat
...
+        Returns:
+            complex: The standard error of this operator with regards to the
+                state preparation.
+        """
+        variance = self.get_variance(qubit_operator, state_prep_circuit, initial_statevector)
+
+        if not self.n_shots:
```
ValentinS4t1qbit, line 343: Python offers its own version of the "ternary if", which is pretty concise and natural to read:
`return np.sqrt(variance/self.n_shots) if self.n_shots else 0.`

tangelo/linq/simulator_base.py
```diff
@@ -231,6 +265,86 @@ def get_expectation_value(self, qubit_operator, state_prep_circuit, initial_stat
...
+        draw samples if required. In the case of a noisy simulator, or if the
+        statevector is not available on the target backend, individual shots
+        must be run and the workflow is akin to what we would expect from an
+        actual QPU.
+
+        Args:
+            qubit_operator(openfermion-style QubitOperator class): a qubit
+                operator.
+            state_prep_circuit: an abstract circuit used for state preparation.
+
```
ValentinS4t1qbit, line 336: `initial_statevector` should be documented

tangelo/linq/simulator_base.py
```diff
@@ -284,6 +398,54 @@ def _get_expectation_value_from_statevector(self, qubit_operator, state_prep_cir
...
+        ket |\psi>. Return the variance <\psi | H | \psi>, computed
+        without drawing samples (statevector only). Users should not be calling
+        this function directly, please call "get_expectation_value" instead.
+
+        Args:
+            qubit_operator(openfermion-style QubitOperator class): a qubit
+                operator.
+            state_prep_circuit: an abstract circuit used for state preparation
+                (only pure states).
+
```
ValentinS4t1qbit, line 412: initial_statevector doc

tangelo/linq/simulator_base.py
```diff
@@ -284,6 +398,54 @@ def _get_expectation_value_from_statevector(self, qubit_operator, state_prep_cir
...
+        variance = 0.
+        prepared_frequencies, prepared_state = self.simulate(state_prep_circuit, return_statevector=True, initial_statevector=initial_statevector)
+
+        # Otherwise, use generic statevector expectation value
+        for term, coef in qubit_operator.terms.items():
+
+            if len(term) > n_qubits:  # Cannot have a qubit index beyond circuit size
+                raise ValueError(f"Size of operator {qubit_operator} beyond circuit width ({n_qubits} qubits)")
+            elif not term:  # Empty term: no simulation needed
+                variance += 0.
```
ValentinS4t1qbit, line 428: I don't understand why this instruction is here. This does nothing, so it does not need to be in the code at all, we can just `continue`.

btw Python has the `pass` keyword for when we do need an instruction that does nothing, for example a class / function that needs at least an instruction so that Python does not complain. If you however have a conditional branch that just says `pass` then it's very likely you don't need that branch at all in the first place.

tangelo/linq/simulator_base.py
```diff
@@ -284,6 +398,54 @@ def _get_expectation_value_from_statevector(self, qubit_operator, state_prep_cir
...
+                continue
+
+            if not self.n_shots:
+                # Without shots, variance using statevector is identically zero
+                variance += 0.
```
ValentinS4t1qbit: If there are no shots we do nothing. So why don't we just test for the opposite case (`if self.n_shots`) and only do what's needed there? Then we can completely delete L431-433 I believe. 

It seems however further conversation may be needed about what this branch should actually do.

tangelo/linq/simulator_base.py
```diff
@@ -326,6 +488,50 @@ def _get_expectation_value_from_frequencies(self, qubit_operator, state_prep_cir
...
+    def _get_variance_from_frequencies(self, qubit_operator, state_prep_circuit, initial_statevector=None):
+        r"""Take as input a qubit operator H and a state preparation returning a
+        ket |\psi>. Return the variance of <\psi | H | \psi> computed
+        using the frequencies of observable states.
+
+        Args:
+            qubit_operator(openfermion-style QubitOperator class): a qubit
+                operator.
+            state_prep_circuit: an abstract circuit used for state preparation.
+
```
ValentinS4t1qbit, line 496: initial_statevector

tangelo/linq/simulator_base.py
```diff
@@ -326,6 +488,50 @@ def _get_expectation_value_from_frequencies(self, qubit_operator, state_prep_cir
...
+            initial_circuit = Circuit(n_qubits=n_qubits)
+            _, updated_statevector = self.simulate(state_prep_circuit, return_statevector=True, initial_statevector=initial_statevector)
+
+        variance = 0.
+        for term, coef in qubit_operator.terms.items():
+
+            if len(term) > n_qubits:
+                raise ValueError(f"Size of operator {qubit_operator} beyond circuit width ({n_qubits} qubits)")
+            elif not term:  # Empty term: no simulation needed
+                variance += 0.
```
ValentinS4t1qbit, line 518: remove this line, it is not necessary

tangelo/linq/tests/test_simulator.py
```diff
@@ -147,6 +147,43 @@ def test_get_exp_value_mixed_state(self):
             results[b] = sim.get_expectation_value(op1, circuit_mixed)
             np.testing.assert_almost_equal(results[b], reference, decimal=2)
 
+    def test_get_variance(self):
```
ValentinS4t1qbit, line 150: I'm not sure I understand the point of testing `get_expectation_value` here.
You are not reusing those values anyway, as we recompute a brand new value in `get_variance` (exact, or from shots).

Maybe this was a little sanity check on your side to help with debugging and development ?
All I currently see is tests that do not test anything that was not tested already, unrelated to your feature, with each a probably to break depending on the samples.

ValentinS4t1qbit: I think we should however test whatever that code is not using (e.g `_get_variance_from_frequencies` or `_get_variance_from_statevector`) explicitly to make sure both the underlying functions you have written are behaving as you'd expect.

tangelo/linq/tests/test_simulator.py
```diff
@@ -147,6 +147,43 @@ def test_get_exp_value_mixed_state(self):
...
+        """ Test variance for simple analytical circuit. """
+
+        opx = 1.0 * QubitOperator("X0")
+        opy = 1.0 * QubitOperator("Y0")
+        opz = 1.0 * QubitOperator("Z0")
+
+        # prepares sqrt(2/3)|0> + -i*sqrt(1/3)|1>
+        circuit = Circuit([Gate("RX", 0, parameter=2*np.arcsin(np.sqrt(1/3)))])
+
+        for b in installed_simulator:
```
ValentinS4t1qbit, line 160: @AlexandreF-1qbit @JamesB-1qbit Since all the variance functions  have been implemented in `SimulatorBase` and rely on the existing and thoroughly tested `get_expectation_value` method each backend implements: is there any meaning in testing this for all simulators ?

Couldn't we just use the default cirq simulator for that test ? We can then maybe crank it up to 10**6 shots to guarantee the comparison up to 2 decimals never fails because of "bad luck" (happens on rare occasions in other tests for me in my recent developments).

tangelo/linq/simulator_base.py
jjgoings: Looks like we didn't need extra logic! Everything is computed the same way with frequencies now, either ideal or re-sampled depending on n_shots.

jjgoings:
@AlexandreF-1qbit should be good now! Tests for both n_shots = None and n_shots = 10**5 should pass.

ValentinS4t1qbit (changes requested):
Thank you for your PR Joshua, really appreciate it ! You did a great job with adopting our coding guidelines and seemed to have found the right place to add your feature.

 I found some minor things, that should be fairly easy to resolve / clean up.

There's code that may be redundant (reused from expectation value functions) but I'm not sure yet whether we should refactor this. For now I think it's better to address what has been brought up and resolve ongoing conversations: then it may become more obvious.

ValentinS4t1qbit:
> @AlexandreF-1qbit should be good now! Tests for both n_shots = None and n_shots = 10**5 should pass.

Dammit, I just gave you a review but it may mean it was given on code that's now outdated. Hopefully the stuff is still relevant

tangelo/linq/tests/test_simulator.py
jjgoings: Since statevector is still based on ideal frequencies, we can handle all the functionality with `from_frequencies`.

jjgoings: And those tests should be gone now :)

tangelo/linq/tests/test_simulator.py
jjgoings: We use just the cirq simulator now.

AlexandreF-1qbit (approved):

jjgoings:
@ValentinS4t1qbit I believe I addressed all your points -- thanks! Let me know if there is anything more I can do.
I wanted to bring up two points for your consideration:

1. Regarding refactoring, I totally agree this code is very similar to the `get_expectation`, as that's what I used as a template. I wonder if in the future we could have a function like `get_moment` where you can pass which [moment](https://en.wikipedia.org/wiki/Moment_(mathematics)) you want to compute for mean, variance, skewness, etc.
2. The functions here for variance assumes that covariance between Pauli products is zero. This is true if each term of a Hamiltonian is measured with an independent circuit. But when you group terms to reduce the number of circuits, this is not necessarily true. Depending on how you group the terms, this can make the variance increase (if Cov(Pa,Pb) > 0) or decrease (Cov(Pa,Pb) < 0). I'm not sure how to handle this, because it seems like a case-by-case basis? One fascinating pathological example was given in Sec B.2 of [this paper](https://arxiv.org/abs/1509.04279) (starts on page 15), also related in Sec IX.A in [this paper](https://par.nsf.gov/servlets/purl/10286356). Turns out that although you can partition the expectation into two measurements, it will still require more shots (33% more) to achieve the same precision as the seemingly sub-optimal three measurement partitioning. Anyway, don't really have a suggestion here other than to be aware and maybe in the future we can support computing variances given a QWC group in post-processing.

ValentinS4t1qbit (approved):
Thank you for contributing this feature to Tangelo Josh, we really appreciate it.
Congratulations on your 1st PR to this project !

PR #244, opened 2022-11-04
Add multi-control support for trotterization

JamesB-1qbit:
The current implementation makes a simplification for when an identity qubit operator is trotterized that is not applicable for multiple controls. This is fixed.


ValentinS4t1qbit (approved):
Thank you !

PR #245, opened 2022-11-07
Deprecate Simulator, introduce get_backend as replacement

ValentinS4t1qbit:
Name change to better reflect the flexibility of the new backend approach in the linq module.

- The `SimulatorBase` class has now been renamed `Backend`. As a consequence, anything that inherits from it is a backend: simulator, or actual quantum device (there is thus no need to change the names of built-in backends, they are indeed simulators).
- Users can now get a backend object by calling `get_backend`. `Simulator` still works but gives a deprecation warning message.
- Minor refactorings


AlexandreF-1qbit (commented):
The changes are good for me. However, the documentation should be changed accordingly everywhere (docstrings, notebooks, comments). Is it a task for this PR? If not, we should raise an issue to keep that in mind. I don't feel it is mandatory as the SImulator class is still supported but comes with a deprecated warning.

For e.g., here are the docstrings and comments that a `grep -r "Simulator" *` raises (excluding the notebooks):
```
tangelo/algorithms/variational/adapt_vqe_solver.py:            backend (tangelo.linq.Simulator): Backend to compute expectation values.
tangelo/algorithms/variational/sa_vqe_solver.py:        backend_options (dict) : parameters to build the tangelo.linq Simulator class.
tangelo/algorithms/variational/iqcc_ilc_solver.py:        backend_options (dict): Parameters to build the tangelo.linq Simulator
tangelo/algorithms/variational/sa_oo_vqe_solver.py:        backend_options (dict) : parameters to build the tangelo.linq Simulator class.
tangelo/algorithms/variational/iqcc_solver.py:        backend_options (dict): Parameters to build the tangelo.linq Simulator
tangelo/algorithms/variational/vqe_solver.py:        backend_options (dict) : parameters to build the tangelo.linq Simulator
tangelo/toolboxes/circuits/tests/test_qsp.py:# Initiate Simulator using cirq for phase estimation tests as it has the same ordering as openfermion
tangelo/toolboxes/circuits/tests/test_lcu.py:# Initiate Simulator using cirq for phase estimation tests as it has the same ordering as openfermion
tangelo/toolboxes/measurements/classical_shadows/classical_shadows.py:            backend (Simulator): Backend for the simulation of a shadow.
tangelo/linq/noisy_simulation/noise_models.py:The Simulator class is responsible for taking in the generic noise model at
tangelo/linq/target/backend.py:"""SimulatorBase class, base class to define various target simulators and abstracting their
tangelo/linq/target/backend.py:differences from the user. Users can define their own target simulator as a child class of SimulatorBase.
tangelo/linq/target/backend.py:                             "Please set the Simulator.n_shots attribute to an appropriate value.")
tangelo/linq/target/target_cirq.py:            # Only DensityMatrixSimulator handles noise well, can use Simulator but it is slower
tangelo/linq/simulator.py:            target_dict (from tangelo.linq.target). Can also provide a subclass of SimulatorBase.
tangelo/linq/simulator.py:    # If target is a string use target_dict to return built-in Target Simulators
tangelo/linq/simulator.py:    # If subclass of SimulatorBase, use target
tangelo/linq/simulator.py:def Simulator(target: Union[None, str, Type[Backend]] = default_simulator, n_shots: Union[None, int] = None,
tangelo/helpers/utils.py:# List all backends and statevector ones supported by Simulator class
```

AlexandreF-1qbit (approved):

JamesB-1qbit (approved):

PR #247, opened 2022-11-08
allow single flip index dis for qcc

JamesB-1qbit:
JKMN mapping requires single flip index generators to converge using QCC. This PR allows a single flip index DIS group where the smallest size before was 2.

### DIFF
--- a/tangelo/toolboxes/ansatz_generator/_qubit_cc.py
+++ b/tangelo/toolboxes/ansatz_generator/_qubit_cc.py
@@ -130,8 +130,8 @@ def get_idxs_deriv(qham_term, *qham_qmf_data):
             gen = (idx, "Y") if idxs == "" else (idx, "X")
             idxs = idxs + f" {idx}" if idxs != "" else f"{idx}"
             gen_tup += (gen, )
-    # Generators must have at least two flip indices
-    if len(gen_tup) > 1:
+    # Generators must have at least one flip index
+    if len(gen_tup) > 0:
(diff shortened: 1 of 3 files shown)


tangelo/toolboxes/ansatz_generator/_qubit_cc.py
```diff
@@ -130,8 +130,8 @@ def get_idxs_deriv(qham_term, *qham_qmf_data):
             gen = (idx, "Y") if idxs == "" else (idx, "X")
             idxs = idxs + f" {idx}" if idxs != "" else f"{idx}"
             gen_tup += (gen, )
-    # Generators must have at least two flip indices
-    if len(gen_tup) > 1:
+    # Generators must have at least one flip index
+    if len(gen_tup) > 0:
```
AlexandreF-1qbit, line 134: Naïve question: is this increasing significantly the pool size, i.e. increase the time to get all the generators per iteration?

JamesB-1qbit: No, it only seems to apply for the JKMN mapping and it is not a significant increase. At most it is N extra generators where N is the number of qubits.

AlexandreF-1qbit: Alright! It is fine

AlexandreF-1qbit (approved):

ValentinS4t1qbit (approved):

PR #250, opened 2022-11-09
Fix link of fig in qchem notebook

AlexandreF-1qbit:
The link was pointing to google drive. This should have been changed to a local link.


ValentinS4t1qbit (approved):
Thank you !

PR #251, opened 2022-11-15
pUCCD ansatz

AlexandreF-1qbit:
Highlights:
- pUCCD ansatz based on V.E. Elfving, M. Millaruelo, J.A. Gámez, and C. Gogolin, Phys. Rev. A 103, 032605 (2021).
- Introduced the Hard-Core Boson (HCB) mapping.
- (CANCELLED) Moved the `assert_term_dict_almost_equal` function in our helper module.
- (Minor) Moved the `spatial_from_spinorb` function. This was to fix a circular import error.


tangelo/algorithms/variational/vqe_solver.py
```diff
@@ -125,6 +126,9 @@ def __init__(self, opt_dict):
                 warnings.warn("Spin-orbital ordering shifted to all spin-up first then down to ensure efficient generator screening "
                               "for the Jordan-Wigner mapping with QCC-based ansatze.", RuntimeWarning)
                 self.up_then_down = True
+            if self.ansatz == BuiltInAnsatze.pUCCD and self.qubit_mapping.lower() != "hcb":
+                warnings.warn("Forcing the hard-core boson mapping for the pUCCD ansatz.", RuntimeWarning)
```
ValentinS4t1qbit, line 130: Is that the official terminology ? Genuinely curious :)

tangelo/helpers/utils.py
```diff
@@ -44,6 +44,22 @@ def assert_freq_dict_almost_equal(d1, d2, atol):
                 raise AssertionError(f"Frequency {k}, difference above tolerance {atol}: {d1[k]} != {d2[k]}")
 
 
+def assert_term_dict_almost_equal(d1, d2, delta=1e-6):
```
ValentinS4t1qbit, line 47: We have a qubit operator class with built-in operations like `__eq__` etc
Does it make sense to create a method for `almost_equal`, that takes a tolerance (and ignores mapping)?

Rather than for people to end up finding this helper by luck, I'd instead extend our class to raise their chance of finding it. It is totally related to qubit operators.


You also could look instead at the absolute value of your highest coefficient of the terms dictionary appearing in the difference of the two qubit operators (difference in mappings may raise an error and prevent the operation).

tangelo/toolboxes/ansatz_generator/ansatz_utils.py
```diff
@@ -425,10 +425,10 @@ def givens_gate(target, theta, is_variational=False):
     """Generates the list of gates corresponding to a givens rotation exp(-theta*(XX+YY))
 
     Explicitly the two-qubit matrix is
-    [[0,      0,           0,       0],
+    [[1,      0,           0,       0],
```
ValentinS4t1qbit, line 428: I can't believe we let that one fly.

tangelo/toolboxes/ansatz_generator/puccd.py
```diff
@@ -0,0 +1,204 @@
...
+
+from tangelo.linq import Circuit
+from tangelo.toolboxes.ansatz_generator import Ansatz
+from tangelo.toolboxes.ansatz_generator.ansatz_utils import givens_gate
+from tangelo.toolboxes.qubit_mappings.statevector_mapping import vector_to_circuit
+
+
+class pUCCD(Ansatz):
+    """This class implements the pUCCD ansatz. Only closed-shell molecules are
+    supported.
```
ValentinS4t1qbit, line 36: I need you to remove the ambiguity

1. If this reflects the current state of the code and open-shell support will be implemented in the future: remove this and let your `NotImplementedError` handle it. We dont want to have to update documentatino every time code change.

2. Otherwise state that pUCCD was not designed to support open-shell systems. Your error then becomes a `ValueError`: the value passed by the users are not adequate. Your current error mean "it's not yet implemented but it will be later"

tangelo/toolboxes/ansatz_generator/puccd.py
```diff
@@ -0,0 +1,204 @@
...
+    Args:
+        molecule (SecondQuantizedMolecule): Self-explanatory.
+        reference_state (string): String refering to an initial state.
+            Default: "HF".
+    """
+
+    def __init__(self, molecule, reference_state="HF"):
+
+        if molecule.spin != 0:
+            raise NotImplementedError("pUCCD is implemented only for closed-shell system.")
```
ValentinS4t1qbit, line 47: NotImplementedError or ValueError ?
is implemented, or is designed ?

tangelo/toolboxes/molecular_computation/rdms.py
```diff
@@ -18,12 +18,13 @@
...
 import numpy as np
 
+from tangelo.toolboxes.molecular_computation.coefficients import spatial_from_spinorb
 from tangelo.linq.helpers import pauli_string_to_of, get_compatible_bases
 from tangelo.toolboxes.operators import FermionOperator
 from tangelo.toolboxes.measurements import ClassicalShadow
 from tangelo.toolboxes.post_processing import Histogram, aggregate_histograms
 from tangelo.toolboxes.qubit_mappings.mapping_transform import fermion_to_qubit_mapping, get_qubit_number
-from tangelo.toolboxes.molecular_computation.molecule import spatial_from_spinorb
+from tangelo.toolboxes.molecular_computation.coefficients import spatial_from_spinorb
```
ValentinS4t1qbit, line 27: duplicate from L21

tangelo/toolboxes/qubit_mappings/hcb.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+def boson_to_qubit_mapping(bos_op):
+    """Function to convert a Bosonic operator to a qubit operator. As qubits are
+    bosons, the mapping is similar to the Jordan-Wigner mapping, but without the
+    trailing Pauli-Z to account for anticommutation of the creation and
+    anhilation operators.
+
+    In short, every creation operator b^{\dagger} (resp. anhilation b) are
+    mapped to X+iY strings (resp. X-iY), where X and Y are refering to the
```
ValentinS4t1qbit, line 63: referring

(this docstring is very well written, I was impressed)

tangelo/toolboxes/qubit_mappings/hcb.py
```diff
@@ -0,0 +1,89 @@
...
+            boson_op += BosonOperator(f"{i}^ {i} {j}^ {j}", r2_coeff)
+
+    return boson_op
+
+
+def boson_to_qubit_mapping(bos_op):
+    """Function to convert a Bosonic operator to a qubit operator. As qubits are
+    bosons, the mapping is similar to the Jordan-Wigner mapping, but without the
+    trailing Pauli-Z to account for anticommutation of the creation and
+    anhilation operators.
```
ValentinS4t1qbit, line 60: anihilation

tangelo/toolboxes/qubit_mappings/hcb.py
```diff
@@ -0,0 +1,89 @@
...
+
+    In short, every creation operator b^{\dagger} (resp. anhilation b) are
```
ValentinS4t1qbit, line 62: anihilation

ValentinS4t1qbit (changes requested):
Rather minor things overall.

Things that end up being helpers should usually be things that do not really have their place somewhere particular and do very general things. If a function pertains to specific objects of features, it should find its place in an existing toolbox somehow. In this situation I think we have a candidate for a new method.

Think about it: if you were looking for such a functionality, where would you intuitively look, as a user ?

tangelo/algorithms/variational/vqe_solver.py
AlexandreF-1qbit: Yep, at least as I saw in the paper. It refers to boson that cannot occupy the same energy levels (so obeying fermionic statistics). The advantage is that we do not have to consider anticommutation relations, i.e. long strings of Z Pauli matrices in the JW mapping.

It is similar to phaseless fermions; the distinction is that we have half the number of particle to consider (2 fermions = 1 boson in this approximation).

tangelo/helpers/utils.py
AlexandreF-1qbit: Thanks for the heads-up,, I will look into this.

tangelo/toolboxes/qubit_mappings/hcb.py
AlexandreF-1qbit: I believe it is annihilation, I had to check it on google.

tangelo/helpers/utils.py
AlexandreF-1qbit: `__eq__` was hard to use, as every digit seems to be checked. I decided to use the `isclose` method instead, with a tight tolerance (https://quantumai.google/reference/python/openfermion/ops/SymbolicOperator#isclose).

tangelo/toolboxes/qubit_mappings/hcb.py
```diff
@@ -0,0 +1,89 @@
...
+operator. For electronic system, it means to couple all the electrons by pair.
+"""
+
+import itertools
+
+from tangelo.toolboxes.operators import BosonOperator
+from tangelo.toolboxes.operators import QubitOperator
+
+
+def hard_core_boson_operator(ferm_op):
```
ValentinS4t1qbit, line 26: you have not documented your arguments

tangelo/toolboxes/qubit_mappings/hcb.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+"""Module that defines function to convert a fermionic operator to a boson
+operator. For electronic system, it means to couple all the electrons by pair.
+"""
+
+import itertools
+
+from tangelo.toolboxes.operators import BosonOperator
+from tangelo.toolboxes.operators import QubitOperator
```
ValentinS4t1qbit, line 23: merge these 2 lines

ValentinS4t1qbit (changes requested):
1 thing left

ValentinS4t1qbit (approved):
Thank you :)

PR #253, opened 2022-11-16
DMET bug with QCC

AlexandreF-1qbit:
DMET was throwing an error when using VQE-QCC. The fix is to trick the `isinstance` function in the QCC ansatz. Also, the `self.solver_fragment_dict` is not defined before simulate when calling `quantum_fragment_data`: the error raised was not helpful. With the fix, the `RuntimeError` with our custom message is printed.

### DIFF
--- a/tangelo/problem_decomposition/dmet/fragment.py
+++ b/tangelo/problem_decomposition/dmet/fragment.py
@@ -20,6 +20,7 @@ import numpy as np
 import pyscf
 from pyscf import ao2mo
 
+from tangelo import SecondQuantizedMolecule
 from tangelo.toolboxes.operators import FermionOperator
 from tangelo.toolboxes.qubit_mappings.mapping_transform import get_fermion_operator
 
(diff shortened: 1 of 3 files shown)


JamesB-1qbit (approved):

tangelo/problem_decomposition/dmet/fragment.py
```diff
@@ -63,6 +64,11 @@ def __post_init__(self):
         self.fermionic_hamiltonian = self._get_fermionic_hamiltonian()
         self.frozen_mos = None
 
+    @property
+    def __class__(self):
+        # Tricking isinstance function.
+        return SecondQuantizedMolecule
```
ValentinS4t1qbit, line 70: I do not like this, because we instantiated a fragment object and now regardless of where it's used, querying for type will return `SecondQuantizedMolecule`. This could lead to sneaky bugs in the future with DMET.

I could not find any reference on when to overwrite the `__class__` method: have you found anything in PEP or in search engines ?

If this object can be considered a `SecondQuantizedMolecule` with a few extra functionalities and fields, then maybe this class should inherit from `SecondQuantizedMolecule` and we could redefine post_init / init. In that situation, we wouldn't even have to fool `isInstance`, as it would return `True`.

It's tempting to just merge, but I just want to have more clarity on the matter. If this is a pain in the butt, well, we may move forward with this solution.

AlexandreF-1qbit: I change the check with the options we discussed via zoom.

ValentinS4t1qbit (approved):
Thank you Alexandre: a simple solution to our problems.

PR #261, opened 2022-12-14
check for small value lcu

JamesB-1qbit:
When multiplying qubit operators, zero coefficients can remain and this was not checked for in the LCU methods which caused nans.

For example
```python3
v1 = QubitOperator((), 1) + QubitOperator("X0", 0.5j)
v2 = QubitOperator((), 1) - QubitOperator("X0", 0.5j)
vdot = v1*v2
print([(term, coeff) for term, coeff in vdot.terms.items()])
```
outputs
```
[((), (1.25+0j)), (((0, 'X'),), 0j)]
```

### DIFF
--- a/tangelo/toolboxes/circuits/lcu.py
+++ b/tangelo/toolboxes/circuits/lcu.py
@@ -322,6 +322,9 @@ def get_uprep_uselect(qu_op: QubitOperator, control: Union[int, List[int]] = Non
         if acoeff > 1.e-8:
             vector += [acoeff]
             unitaries += [QubitOperator(term, -coeff / acoeff)]
+        else:
+            vector += [0]
+            unitaries += [QubitOperator((), 1)]
 
     # create U_{prep} from sqrt of coefficients
     vector = np.array(vector)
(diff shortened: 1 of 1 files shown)


AlexandreF-1qbit (approved):

tangelo/toolboxes/circuits/lcu.py
```diff
@@ -381,7 +382,7 @@ def get_lcu_qubit_op_info(qu_op: QubitOperator()) -> Tuple[List[int], List[int],
     """
 
     max_qu_op = count_qubits(qu_op)
-    num_terms = len(qu_op.terms)
+    num_terms = len([coeff for coeff in qu_op.terms.values() if abs(coeff) > 1.e-8])
```
ValentinS4t1qbit, line 385: If some coefficients are very close to 0, `num_terms` will then not match `len(qu_op.terms)`, which seems a bit misleading to me, and can cause users to have mismatch and unexpected differences when manipulating term-related info. If these terms "do not matter", how about we just remove them ?

I think maybe a good approach could be to set a parameter `threshold` with default value `1.e-8`, that the user can override, and anything below the threshold in absolute value vanishes entirely. 

I don't want to make this a request for change without getting clarity from you, but I'm not ready to approve this either.

ValentinS4t1qbit (commented):
Comment submitted, let's clarify before moving forward. We can rope others into the conversation to make a case or help us decide.

ValentinS4t1qbit (approved):

PR #150, opened 2022-05-02
Extended mixed state

JamesB-1qbit:
Save mid-circuit measurements for cirq, qiskit, qdk, and qulacs.
desired_meas_result permits users to select which measurement values they want to post-select. The frequency dictionary and statevector(possibly density matrix for cirq) returned corresponds to the requested mid-circuit measurement.

PR #263, opened 2023-01-09
Desired measurement result

KrzysztofB-1qbit:
Adds functionality to post-select samples by comparing mid-circuit measurement results to a user-provided desired measurement bitstring.

Closes #150 (all features there are included here)


tangelo/linq/target/target_qulacs.py
```diff
@@ -74,9 +77,9 @@ def simulate_circuit(self, source_circuit: Circuit, return_statevector=False, in
             state.load(initial_statevector)
 
         # If you don't want to save the mid-circuit measurements for a mixed state
-        if (source_circuit.is_mixed_state or self._noise_model) and not save_mid_circuit_meas:
+        if (source_circuit.is_mixed_state or self._noise_model) and (desired_meas_result is None and not save_mid_circuit_meas):
             samples = dict()
-            for _ in range(self.n_shots):
+            for i in range(self.n_shots):
```
JamesB-1qbit, line 82: `i` does not appear to be used.

tangelo/linq/target/backend.py
```diff
@@ -205,10 +207,18 @@ def simulate(self, source_circuit, return_statevector=False, initial_statevector
...
         """
+        n_meas = source_circuit.counts.get("MEASURE", 0)
+
         if source_circuit.is_mixed_state and not self.n_shots:
             raise ValueError("Circuit contains MEASURE instruction, and is assumed to prepare a mixed state."
                              "Please set the n_shots attribute to an appropriate value.")
 
+        if desired_meas_result is not None:
+            if len(desired_meas_result) != n_meas or not isinstance(desired_meas_result, str):
+                raise ValueError("desired_meas result is not a string with the same length as the number of measurements"
```
JamesB-1qbit, line 218: This should be 
```python3
if not isinstance(desired_meas_result, str) or len(desired_meas_result) != n_meas:
     raise ValueError(f"desired_meas_result of {desired_meas_result} is not a string with the same length as the number of measurements = {n_meas}"
```
as it will throw an error if someone inputs an object that does not have have a `len`.

You may also want a test of this in `test_simulator`

JamesB-1qbit (changes requested):
A couple small things.

tangelo/linq/target/backend.py
```diff
@@ -224,24 +234,29 @@ def simulate(self, source_circuit, return_statevector=False, initial_statevector
                 statevector[0] = 1.0
             return (frequencies, statevector) if return_statevector else (frequencies, None)
 
+        # For mid-circuit measurements post-process the result
         if save_mid_circuit_meas:
             # TODO: refactor to break a circular import. May involve by relocating get_xxx_oneterm functions
```
AlexandreF-1qbit, line 239: Is the circular import still there?

tangelo/linq/target/target_qiskit.py
```diff
@@ -92,11 +96,34 @@ def simulate_circuit(self, source_circuit: Circuit, return_statevector=False, in
...
+            qiskit_noise_model = get_qiskit_noise_model(self._noise_model) if self._noise_model else None
+            translated_circuit = self.qiskit.transpile(translated_circuit, backend)
+            translated_circuit.save_statevector()
+            samples = dict()
+            self._current_state = None
+
+            for _ in range(self.n_shots):
+                sim_results = backend.run(translated_circuit, noise_model=qiskit_noise_model, shots=1).result()
+                current_state = sim_results.get_statevector(translated_circuit)
+                measurement = next(iter(self.qiskit.result.marginal_counts(sim_results, indices=list(range(n_meas))).get_counts()))[::-1]
```
AlexandreF-1qbit, line 118: One question here: how the `marginal_counts` function play with the `list(range(n_meas))`? With this code:
https://github.com/goodchemistryco/Tangelo/blob/3f354fe35c0a7e5ac9dc57f9bfaa124d06ad5b73/tangelo/linq/target/backend.py#L247-L249
I understand that the first `n_meas` bits in the bitstring are for the mid-circuit measurements. In the code above, as qiskit qubit ordering is reversed, should the indices be a list of `n_meas` integers from the last qubit index instead of `list(range(n_meas)))`?

tangelo/linq/target/target_qulacs.py
```diff
@@ -95,10 +98,15 @@ def simulate_circuit(self, source_circuit: Circuit, return_statevector=False, in
                 sample = self._int_to_binstr(state.sampling(1)[0], source_circuit.width)
                 bitstr = measurement + sample
                 samples[bitstr] = samples.get(bitstr, 0) + 1
+
+                if measurement == desired_meas_result:
+                    python_statevector = state.get_vector()
+                    self._current_state = python_statevector
```
AlexandreF-1qbit, line 104: Why not
```python
self._current_state =  state.get_vector()
```

tangelo/linq/tests/test_simulator.py
```diff
@@ -139,6 +139,20 @@ def test_simulate_mixed_state_save_measures(self):
             assert_freq_dict_almost_equal(sim.all_frequencies, reference_all, 8e-2)
             assert_freq_dict_almost_equal(sim.mid_circuit_meas_freqs, reference_mid, 8e-2)
 
+    def test_simulate_mixed_state_desired_state(self):
+        """ Test mid-circuit measurement (mixed-state simulation) for compatible/testable formats and backends.
+        Mixed-state do not have a statevector representation, as they are a statistical mixture of several statevectors.
+        Simulating individual shots is suitable.
```
AlexandreF-1qbit, line 145: Mixed-state -> Mixed states, they are a statistical mixture -> they are statistical mixtures
+
What does mean suitable in that case? Is it possible but not recommended?

AlexandreF-1qbit: (same comments for below)

tangelo/linq/tests/test_simulator.py
```diff
@@ -362,6 +392,18 @@ def test_get_exp_value_from_statevector_with_shots_h2(self):
         energy = simulator.get_expectation_value(qubit_operator, abs_circ)
         self.assertAlmostEqual(energy, expected, delta=1e-3)
 
+    def test_get_exp_value_mixed_state_desired_measurement_with_shots(self):
+        """ Get expectation value of mixed state post-selecting on desired measurement"""
```
AlexandreF-1qbit, line 396: post-selecting -> postselection? post-selection?

AlexandreF-1qbit (changes requested):
Some small comments. Also, 
- the `backend.success_probability` is only defined when a `desired_meas_result` is provided, right?
- Are all the frequencies before maginalization stored somewhere?

JamesB-1qbit (commented):
One thing that is missing that I missed in the previous PR is that a statevector is not returned if requested. To make it compatible with the `desired_meas_result`. Only change `self._current_state` to a result that is not `desired_meas_result` if `self._current_state` is none. For example, for `target_cirq`
```python3
if measure == desired_meas_result or self._current_state is None:
     self._current_state = current_state
```
This is useful for work with Isaac Kim in that we need the statevector output every time but our next operation depends on the measurement so we don't have a `desired_meas_result`.

tangelo/linq/target/target_qulacs.py
KrzysztofB-1qbit: ok

tangelo/linq/target/backend.py
KrzysztofB-1qbit: ok

tangelo/linq/target/backend.py
KrzysztofB-1qbit: I think so. Right now it's worked around by importing the offending module locally. Fixing it is not really directly related to this PR but rather design choices in tangelo.

tangelo/linq/target/target_qiskit.py
KrzysztofB-1qbit: I think since we're using the `qiskit.result.marginal_counts` functionality here to return results, they should be the right qubits. Afterwards, they are reversed with `[::-1]` to ensure the right ordering.

tangelo/linq/target/target_qulacs.py
KrzysztofB-1qbit: It's just for consistency, so the `return` statement always uses this variable.

tangelo/linq/tests/test_simulator.py
KrzysztofB-1qbit: Updated the docstring to be in line with the previous test

tangelo/linq/tests/test_simulator.py
KrzysztofB-1qbit: "by post-selecting"

JamesB-1qbit (approved):

tangelo/linq/target/target_qiskit.py
```diff
@@ -92,11 +96,34 @@ def simulate_circuit(self, source_circuit: Circuit, return_statevector=False, in
             job_sim = self.qiskit.execute(translated_circuit, backend, noise_model=qiskit_noise_model,
                                           shots=self.n_shots, basis_gates=None, optimization_level=opt_level)
             sim_results = job_sim.result()
-            frequencies = {state[::-1]: count/self.n_shots for state, count in sim_results.get_counts(0).items()}
+            self.all_frequencies = {state[::-1]: count/self.n_shots for state, count in sim_results.get_counts(0).items()}
 
-            self.all_frequencies = frequencies.copy()
+            frequencies = self.all_frequencies
```
ValentinS4t1qbit, line 101: Attention: here `frequencies` and `self.all_frequencies` are now pointing to the same location in memory. Changing one will change the other. In the previous version of the code, for some reason, one was pointing to a copy and thus changing one would not change the other in the subsequent lines.

Are we sure this is the desired behaviour ?

tangelo/linq/target/target_qiskit.py
```diff
@@ -92,11 +96,34 @@ def simulate_circuit(self, source_circuit: Circuit, return_statevector=False, in
...
             self._current_state = None
 
+        # desired_meas_result is not None and return_statevector is requested so loop shot by shot (much slower)
+        elif desired_meas_result is not None:
+            from tangelo.linq.noisy_simulation.noise_models import get_qiskit_noise_model
+            n_meas = source_circuit.counts.get("MEASURE", 0)
```
ValentinS4t1qbit, line 107: Can't this import be at the beginning of the file ?

tangelo/linq/target/target_qiskit.py
```diff
@@ -92,11 +96,34 @@ def simulate_circuit(self, source_circuit: Circuit, return_statevector=False, in
...
+            self._current_state = None
+
+            for _ in range(self.n_shots):
+                sim_results = backend.run(translated_circuit, noise_model=qiskit_noise_model, shots=1).result()
+                current_state = sim_results.get_statevector(translated_circuit)
+                measure = next(iter(self.qiskit.result.marginal_counts(sim_results, indices=list(range(n_meas))).get_counts()))[::-1]
+                (sample, _) = self.qiskit.quantum_info.states.Statevector(current_state).measure()
+                bitstr = measure + sample[::-1]
+                samples[bitstr] = samples.get(bitstr, 0) + 1
+                if measure == desired_meas_result or self._current_state is None:
```
ValentinS4t1qbit, line 122: I do not understand this either (see similar comment above).
I'm not sure why L105 is useful as a result.

tangelo/linq/target/target_qiskit.py
```diff
@@ -92,11 +96,34 @@ def simulate_circuit(self, source_circuit: Circuit, return_statevector=False, in
...
                                           shots=self.n_shots, basis_gates=None, optimization_level=opt_level)
             sim_results = job_sim.result()
-            frequencies = {state[::-1]: count/self.n_shots for state, count in sim_results.get_counts(0).items()}
+            self.all_frequencies = {state[::-1]: count/self.n_shots for state, count in sim_results.get_counts(0).items()}
 
-            self.all_frequencies = frequencies.copy()
+            frequencies = self.all_frequencies
             self._current_state = None
 
+        # desired_meas_result is not None and return_statevector is requested so loop shot by shot (much slower)
```
ValentinS4t1qbit, line 104: Discrepancy between L104 and 105 ? Does it even make sense to return a state representation unless its a density matrix, for a mixed state ?

tangelo/linq/target/target_qulacs.py
ValentinS4t1qbit: does this even make sense? Like, if this "current state" is not the same for all shots and thus not deterministic, what meaning does it even have ?

As a user, what do I do with that? I won't guess it unless I look at the code, and I'll understand (or not) that t's just the last state that apparently matched my desired mid-circuit measurement. May be misleading.

tangelo/linq/tests/test_simulator.py
```diff
@@ -139,6 +139,25 @@ def test_simulate_mixed_state_save_measures(self):
...
+        results = dict()
+        exact = {'11': 0.23046888414227926, '10': 0.7695311158577207}
+        for b in installed_simulator:
+            sim = get_backend(target=b, n_shots=10 ** 3)
+            results[b], _ = sim.simulate(circuit_mixed, desired_meas_result="0")
+            assert_freq_dict_almost_equal(results[b], exact, 8.e-2)
+
+    def test_desired_meas_len(self):
+        """ Test if the desired_meas_result parameter is a string and of the right length."""
+        for b in installed_simulator:
```
ValentinS4t1qbit, line 156: This test does not depend on the choice of backend. This loop is unnecessary.

tangelo/linq/tests/test_simulator.py
```diff
@@ -139,6 +139,25 @@ def test_simulate_mixed_state_save_measures(self):
...
             assert_freq_dict_almost_equal(sim.mid_circuit_meas_freqs, reference_mid, 8e-2)
 
+    def test_simulate_mixed_state_desired_state(self):
+        """ Test mid-circuit measurement (mixed-state simulation) for all installed backends.
+        Mixed-states do not have a statevector representation, as they are statistical mixtures of several quantum states.
+        """
+
+        results = dict()
+        exact = {'11': 0.23046888414227926, '10': 0.7695311158577207}
+        for b in installed_simulator:
```
ValentinS4t1qbit, line 149: Doesn't feel like you implemented anything for QDK unless I'm wrong. If you only implemented this feature for a few specific backends, list them explicitly.

tangelo/linq/tests/test_simulator.py
```diff
@@ -139,6 +139,25 @@ def test_simulate_mixed_state_save_measures(self):
             assert_freq_dict_almost_equal(sim.all_frequencies, reference_all, 8e-2)
             assert_freq_dict_almost_equal(sim.mid_circuit_meas_freqs, reference_mid, 8e-2)
 
+    def test_simulate_mixed_state_desired_state(self):
+        """ Test mid-circuit measurement (mixed-state simulation) for all installed backends.
+        Mixed-states do not have a statevector representation, as they are statistical mixtures of several quantum states.
```
ValentinS4t1qbit, line 144: Mixed states.

Is this information relevant to the test ? Mixed states have a density matrix representation I believe, but I also am not sure why we'd mention any of that here.

tangelo/linq/tests/test_simulator.py
```diff
@@ -205,6 +224,22 @@ def test_simulate_statevector(self):
                 frequencies, _ = simulator.simulate(circuit)
                 assert_freq_dict_almost_equal(ref_freqs[i], frequencies, atol=1e-5)
 
+    def test_simulate_mixed_state_desired_statevector(self):
+        """ Test mid-circuit measurement (mixed-state simulation) for compatible/testable formats and backends.
+        Mixed-state do not have a statevector representation, as they are a statistical mixture of several statevectors.
```
ValentinS4t1qbit, line 229: same as above

tangelo/linq/tests/test_simulator.py
```diff
@@ -461,7 +508,8 @@ def __init__(self, n_shots=None, noise_model=None, return_zeros=True):
                 super().__init__(n_shots=n_shots, noise_model=noise_model)
                 self.return_zeros = return_zeros
 
-            def simulate_circuit(self, source_circuit: Circuit, return_statevector=False, initial_statevector=None, save_mid_circuit_meas=False):
+            def simulate_circuit(self, source_circuit: Circuit, return_statevector=False, initial_statevector=None,
+                                 desired_meas_result=None, save_mid_circuit_meas=False):
```
ValentinS4t1qbit, line 512: Was this returning an error because of inheritance ? I thought python wouldn't bother us with signature match.

Here `initial_statevector=None, desired_meas_result=None, save_mid_circuit_meas=False` are not options supported by the backend and can be removed, unless python complains because of inheritance.

tangelo/linq/tests/test_simulator_noisy.py
```diff
@@ -163,6 +167,10 @@ def test_noisy_simulation_qiskit(self):
         res_mixed, _ = s_nmm.simulate(circuit_mixed)
         assert_freq_dict_almost_equal(ref_mixed, res_mixed, 7.e-2)
 
+        s_nmm = get_backend(target="qiskit", n_shots=10**4, noise_model=nmm)
```
ValentinS4t1qbit, line 170: I need a comment to describe what this tests if you're lumping it with other things

tangelo/linq/tests/test_simulator_noisy.py
```diff
@@ -194,6 +202,20 @@ def test_noisy_simulation_cirq(self):
         res_mixed, _ = s_nmm.simulate(circuit_mixed)
         assert_freq_dict_almost_equal(ref_mixed, res_mixed, 7.e-2)
 
+        s_nmm = get_backend(target="cirq", n_shots=10**4, noise_model=nmm)
```
ValentinS4t1qbit, line 205: Same as above

ValentinS4t1qbit (changes requested):
A number of things seem unclear to me (linq is a complicated beast to change).

I wonder if this post-selection should be handled by simulate, or done separately. Handling it with simulate sure feels somewhat convenient but is entangling quantum execution from post-processing. This code is already tricky and complex, we should be very careful about our design choices.

tangelo/linq/target/backend.py
```diff
@@ -224,24 +234,29 @@ def simulate(self, source_circuit, return_statevector=False, initial_statevector
                 statevector[0] = 1.0
             return (frequencies, statevector) if return_statevector else (frequencies, None)
 
+        # For mid-circuit measurements post-process the result
```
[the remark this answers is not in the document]
AlexandreF-1qbit: I kind of understand the point, but this way would prevent (if I understood well) some very useful features:
- Classically controlled gates (the mid-circuit measurements are not useful if stored in memory).
- Reusing qubits. We can circumvent that by adding other qubits, but emulating this feature becomes difficult to do in the case where many RESET are needed.

tangelo/linq/target/backend.py
```diff
@@ -224,24 +234,29 @@ def simulate(self, source_circuit, return_statevector=False, initial_statevector
...
         if save_mid_circuit_meas:
             # TODO: refactor to break a circular import. May involve by relocating get_xxx_oneterm functions
             from tangelo.toolboxes.post_processing.post_selection import split_frequency_dict
 
             (all_frequencies, statevector) = self.simulate_circuit(source_circuit,
```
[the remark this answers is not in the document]
KrzysztofB-1qbit: The variable `all_frequencies` includes both the results of the mid-circuit measurements and the final measurements. This is then split in L247 and post-processed if `desired_meas` was provided. The mid-circuit ones are then stored in `self.mid_circuit_meas_freqs` and the final measurements are returned.

tangelo/linq/target/backend.py
```diff
@@ -224,24 +234,29 @@ def simulate(self, source_circuit, return_statevector=False, initial_statevector
...
                                                                    return_statevector=return_statevector,
                                                                    initial_statevector=initial_statevector,
+                                                                   desired_meas_result=desired_meas_result,
                                                                    save_mid_circuit_meas=save_mid_circuit_meas)
-            n_meas = source_circuit.counts.get("MEASURE", 0)
-            self.mid_circuit_meas_freqs, frequencies = split_frequency_dict(all_frequencies, list(range(n_meas)))
+            self.mid_circuit_meas_freqs, frequencies = split_frequency_dict(all_frequencies,
```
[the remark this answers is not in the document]
KrzysztofB-1qbit: I think there was a design change at some point. Possibly there was some confusion in one of the iterations. It looks like there's no need for `desired_meas_result` in the `target_*` modules now.

tangelo/linq/target/backend.py
```diff
@@ -175,7 +175,7 @@ def simulate_circuit(self):
         """
         pass
 
-    def simulate(self, source_circuit, return_statevector=False, initial_statevector=None, save_mid_circuit_meas=False):
+    def simulate(self, source_circuit, return_statevector=False, initial_statevector=None, desired_meas_result=None, save_mid_circuit_meas=False):
```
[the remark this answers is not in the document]
KrzysztofB-1qbit: Maybe `desired_mid_circ_result` then?

tangelo/linq/target/backend.py
```diff
@@ -380,6 +398,8 @@ def _get_expectation_value_from_statevector(self, qubit_operator, state_prep_cir
             qubit_operator (openfermion-style QubitOperator class): a qubit operator.
             state_prep_circuit (Circuit): an abstract circuit used for state preparation (only pure states).
             initial_statevector (array): The initial state of the system
+            desired_meas_result (str): The expectation value is taken for over the frequencies
```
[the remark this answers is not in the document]
ValentinS4t1qbit: we established that this is ok, lets use that parameter in the body of the function and take it as input

tangelo/linq/target/backend.py
```diff
@@ -224,24 +234,29 @@ def simulate(self, source_circuit, return_statevector=False, initial_statevector
...
                                                                    return_statevector=return_statevector,
                                                                    initial_statevector=initial_statevector,
+                                                                   desired_meas_result=desired_meas_result,
                                                                    save_mid_circuit_meas=save_mid_circuit_meas)
-            n_meas = source_circuit.counts.get("MEASURE", 0)
-            self.mid_circuit_meas_freqs, frequencies = split_frequency_dict(all_frequencies, list(range(n_meas)))
+            self.mid_circuit_meas_freqs, frequencies = split_frequency_dict(all_frequencies,
+                                                                            list(range(n_meas)),
+                                                                            desired_measurement=desired_meas_result)
+            self.success_probability = self.mid_circuit_meas_freqs.get(desired_meas_result, 0)
```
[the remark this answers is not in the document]
KrzysztofB-1qbit: Remove success_probability for now

tangelo/linq/target/target_cirq.py
```diff
@@ -122,7 +123,9 @@ def simulate_circuit(self, source_circuit: Circuit, return_statevector=False, in
                 sample = "".join([str(int(q)) for q in isamples[0]])
                 bitstr = measure + sample
                 samples[bitstr] = samples.get(bitstr, 0) + 1
-            self.all_frequencies = {k: v / self.n_shots for k, v in sample.items()}
+                if measure == desired_meas_result or self._current_state is None:
+                    self._current_state = current_state
```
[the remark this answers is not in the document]
KrzysztofB-1qbit: Fixed

tangelo/linq/target/target_qiskit.py
KrzysztofB-1qbit: Try move to top if possible

tangelo/linq/tests/test_simulator.py
KrzysztofB-1qbit: use default backend from other tests (it's cirq)

tangelo/linq/tests/test_simulator.py
KrzysztofB-1qbit: Remove QDK from list of targets.

tangelo/linq/target/target_qdk.py
```diff
@@ -24,7 +24,8 @@ def __init__(self, n_shots=None, noise_model=None):
         super().__init__(n_shots=n_shots, noise_model=noise_model)
         self.qsharp = qsharp
 
-    def simulate_circuit(self, source_circuit: Circuit, return_statevector=False, initial_statevector=None, save_mid_circuit_meas=False):
+    def simulate_circuit(self, source_circuit: Circuit, return_statevector=False, initial_statevector=None,
+                         desired_meas_result=None, save_mid_circuit_meas=False):
```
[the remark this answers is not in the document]
KrzysztofB-1qbit: Remove options from QDK if python doesn't complain, or return `NotImplementedError` if it has to be kept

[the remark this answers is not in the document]
KrzysztofB-1qbit: Actually, there is simple handling of mid-circuit measurements in qdk, there is an extension of the register, and desired measurement is handled by the backend module.

tangelo/linq/target/backend.py
```diff
@@ -422,7 +442,7 @@ def _get_expectation_value_from_statevector(self, qubit_operator, state_prep_cir
 
         return expectation_value
 
-    def _get_expectation_value_from_frequencies(self, qubit_operator, state_prep_circuit, initial_statevector=None):
+    def _get_expectation_value_from_frequencies(self, qubit_operator, state_prep_circuit, initial_statevector=None, desired_meas_result=None):
```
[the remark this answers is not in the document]
KrzysztofB-1qbit: ok

tangelo/linq/target/target_cirq.py
```diff
@@ -36,7 +36,8 @@ def __init__(self, n_shots=None, noise_model=None):
         super().__init__(n_shots=n_shots, noise_model=noise_model)
         self.cirq = cirq
 
-    def simulate_circuit(self, source_circuit: Circuit, return_statevector=False, initial_statevector=None, save_mid_circuit_meas=False):
+    def simulate_circuit(self, source_circuit: Circuit, return_statevector=False, initial_statevector=None,
+                         desired_meas_result=None, save_mid_circuit_meas=False):
```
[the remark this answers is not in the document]
KrzysztofB-1qbit: ok

tangelo/linq/tests/test_simulator.py
KrzysztofB-1qbit: ok

tangelo/linq/tests/test_simulator.py
KrzysztofB-1qbit: ok

tangelo/linq/tests/test_simulator_noisy.py
KrzysztofB-1qbit: It's just another noise model. Fixed

tangelo/linq/tests/test_simulator_noisy.py
KrzysztofB-1qbit: Fixed

tangelo/linq/tests/test_simulator.py
KrzysztofB-1qbit: Somehow it still needs `initial_statevector=None`, but the mid-circuit stuff can be removed

tangelo/linq/target/backend.py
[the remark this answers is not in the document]
KrzysztofB-1qbit: I think we agreed to leave as is

ValentinS4t1qbit:
At this point:

- most comments addresses.
- refactor of cirq target done. But qiskit and qulacs need to be rewritten the same way (find repo of the state, sample n_shots from it directly)
- small cleanup things, verify tests are consistent and passing

ValentinS4t1qbit:
Notebooks have been removed from the main repo as per the latest PR. There is therefore a conflict here: we should simply not have anything that tests notebooks.

tangelo/linq/target/target_qiskit.py
```diff
@@ -75,10 +80,57 @@ def simulate_circuit(self, source_circuit: Circuit, return_statevector=False, in
...
+                        else:
+                            freqs = self._statevector_to_frequencies(np.array(current_state))
+                            self.all_frequencies = {measure + meas: val for meas, val in freqs.items()}
+                        return (self.all_frequencies, np.array(self._current_state)) if return_statevector else (self.all_frequencies, None)
+                    else:
+                        (sample, _) = current_state.measure()
+                        bitstr = measure + sample[::-1]
+                        samples[bitstr] = samples.get(bitstr, 0) + 1
+                        n_success += 1
+                elif desired_meas_result is None:
```
ValentinS4t1qbit, line 120: It feels like control flow here may be more complicated that it could be. I'm not sure I understand what's going on. If there are no desired measurement, then why are we inside this complicated while loop in the first place ?
Yet it's inside the branch led by L86 so that means it cannot be `None` anyway ?

I suggest you separate and organize the different cases clearly because `linq` is becoming tough to handle. You are free to create more functions if that helps. Skip lines, use comments, whatever: let's try to make this complicated logic as accessible as possible. It's ok if some lines are redundant, the goal is to end up with a code that's readable and easier to maintain.

tangelo/linq/target/target_qulacs.py
```diff
@@ -74,36 +77,85 @@ def simulate_circuit(self, source_circuit: Circuit, return_statevector=False, in
...
 
-        # If you don't want to save the mid-circuit measurements for a mixed state
-        if (source_circuit.is_mixed_state or self._noise_model) and not save_mid_circuit_meas:
+        if desired_meas_result is not None:
+            n_meas = source_circuit.counts.get("MEASURE", 0)
+            self._current_state = None
+            n_attempts = 0
+            n_success = 0
+            n_shots = self.n_shots if self.n_shots is not None else -1
+            # Permit 0.1% probability events
```
ValentinS4t1qbit, line 86: Skip a line.

Don't hesitate to throw a comment or two to avoid ending up with monolithic code, or skip a line somewhere if you think it helps parsing the main conditional branches. This applies to everything in this file.

tangelo/linq/target/target_qulacs.py
```diff
@@ -74,36 +77,85 @@ def simulate_circuit(self, source_circuit: Circuit, return_statevector=False, in
...
                 if initial_statevector is not None:
                     state.load(initial_statevector)
                 else:
                     state.set_zero_state()
+
             self.all_frequencies = {k: v / self.n_shots for k, v in samples.items()}
             return (self.all_frequencies, python_statevector) if return_statevector else (self.all_frequencies, None)
 
-        # All other cases for shot-based simulation
+        # If saving mid-circuit measurements is not necessary or a noise model
```
ValentinS4t1qbit, line 146: I strongly suggest you start with the easy cases in your file, or what users are most likely to use.

What I personally dislike most is when I'm looking for something simple and the software makes me read about a bunch of complicated things that were not necessary. This is an open-source projects so many people may read or try to modify this code.

tangelo/linq/tests/test_simulator.py
```diff
@@ -22,7 +22,7 @@
 import time
```
ValentinS4t1qbit, line 22: skip a line. not a 3rd party lib.

tangelo/linq/tests/test_simulator.py
```diff
@@ -205,6 +214,51 @@ def test_simulate_statevector(self):
...
+        """ Test mid-circuit measurement (mixed-state simulation) for compatible/testable formats and backends when returning
+        a statevector."""
+
+        results = dict()
+        results["qulacs"] = np.array([0. + 0.j, 0.87758256 + 0.j, 0. + 0.j, -0.47942554 + 0.j])
+        results["qiskit"] = np.array([0. + 0.j, 0.87758256 + 0.j, 0. + 0.j, -0.47942554 + 0.j])
+        results["cirq"] = np.array([0. + 0.j, 0. + 0.j, 0.87758256 + 0.j, -0.47942554 + 0.j])
+        initial_state = np.array([0, 0, 0, 1])
+        freqs_exact = {'10': 0.7701511529340699, '11': 0.2298488470659301}
+        for b in installed_sv_simulator:
```
ValentinS4t1qbit, line 227: skip a line before the loop

ValentinS4t1qbit (changes requested):
Good stuff !

I think my comments are mostly about readability, and trying to keep our complex `linq` logic as understandable as possible. It's not fun, but please have a look at some of the `simulate_circuit`  methods you've changed and think "what kind of code do I want to leave to future me and whoever may read that?"

The more people can understand this code, the more your burden to maintain it yourself decreases.

AlexandreF-1qbit (approved):

ValentinS4t1qbit (changes requested):
We're almost there, good job. At this point I think our code can't get much simpler, but we should:

- systematically label with a comment the conditional branches that correspond to different situation, for the simulation
- refine the flow for the qiskit file (definite your utility functions first, then define fundamental variables, then handle initial statevector, then handle the various case of simulation based on user input).

If we can do that, I think we have a code we can all more comfortably maintain and extend in the future, and I'd be happy to merge.

tangelo/linq/target/target_qiskit.py
```diff
@@ -75,37 +80,119 @@ def simulate_circuit(self, source_circuit: Circuit, return_statevector=False, in
                 initial_state_circuit.initialize(initial_statevector, list(range(n_qubits)))
                 translated_circuit = initial_state_circuit.compose(translated_circuit)
 
-        # Drawing individual shots with the qasm simulator, for noisy simulation or simulating mixed states
-        if self._noise_model or source_circuit.is_mixed_state and not return_statevector:
-            from tangelo.linq.noisy_simulation.noise_models import get_qiskit_noise_model
+        def aer_backend_with_statevector(translated_circuit):
```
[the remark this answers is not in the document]
JamesB-1qbit: I moved the functions up. `n_meas` and `qiskit_noise_model` are required to define `run_and_measure_one_shot`.

tangelo/linq/target/target_qiskit.py
```diff
@@ -75,37 +80,119 @@ def simulate_circuit(self, source_circuit: Circuit, return_statevector=False, in
...
+                        self.all_frequencies = {measure + meas: val for meas, val in freqs.items()}
+
+                n_attempts += 1
+
+            if n_attempts == max_attempts:
+                raise ValueError(f"desired_meas_result was not measured after {n_attempts} attempts")
+
+            frequencies = self.all_frequencies.copy()
+
+        elif save_mid_circuit_meas and self.n_shots in [1]:
```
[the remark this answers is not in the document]
JamesB-1qbit: If someone wants to do a classically controlled gate at this time, they would use this.
```python3
sim = get_backend('qiskit', n_shots=1)
circuit = Circuit([Gate("H", 0), Gate("MEASURE", 0)])
_, sv = sim.simulate(circuit, return_statevector=True, save_mid_circuit_meas=True)
if next(iter(sim.mid_circuit_meas_freqs)) == '1':
    _, sv = sim.simulate(circuit1, return_statevector=True, initial_statevector=sv, save_mid_circuit_meas=True)
else:
    _, sv = sim.simulate(circuit0, return_statevector=True, initial_statevector=sv, save_mid_circuit_meas=True)
```

where `circuit1` is the circuit applied if `1` is measured and `circuit0` is applied if `0` is measured.

ValentinS4t1qbit (approved):
Thank you for providing this new feature and taking it to the finish line !

PR #266, opened 2023-01-16
added draw method to circuit

JamesB-1qbit:
Uses cirq's SVGCircuit such that the following code
```python3
from tangelo.linq import Circuit, Gate

circ = Circuit([Gate("CRZ", target=0, control=[1,2], parameter=1),
                Gate("CNOT", target=0, control=1),
                Gate("RZ", target=0, parameter=0.3)])
circ.draw()
```
makes
[output.svg.pdf](https://github.com/goodchemistryco/Tangelo/files/10427016/output.svg.pdf)

### DIFF
--- a/tangelo/linq/circuit.py
+++ b/tangelo/linq/circuit.py
@@ -20,7 +20,6 @@ characteristics (width, size ...).
 
 import copy
 from typing import List
-import warnings
 
 import numpy as np
 from cirq.contrib.svg import SVGCircuit
@@ -150,7 +149,6 @@ class Circuit:
         """Method to output a prettier version of the circuit for use in jupyter notebooks that uses cirq SVGCircuit"""
         # circular import
         from tangelo.linq.translator.translate_cirq import translate_c_to_cirq
-        warnings.filterwarnings("ignore", message=".*findfont.*")
         cirq_circ = translate_c_to_cirq(self)
         # Remove identity gates that are added in translate_c_to_cirq (to ensure all qubits are initialized) before drawing.
         cirq_circ.__delitem__(0)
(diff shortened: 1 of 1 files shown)


ValentinS4t1qbit:
Thank you for this simple solution, James. Since Cirq is a requirement of Openfermion, itself a requirement of our package, I think this is a very reasonable default solution.

When I try to call the draw method in the basics notebook, I get a bunch of warning for a font not found, in my Ubuntu environment. I think 1 warning line per gate, which will quickly get unbearable. Is there any way we can silence those warnings or not having them come up in the first place ?

Screenshot from 2023-01-17 17-43-39

JamesB-1qbit:
> 

I don't see these warnings and my virtual box seems to be broken. But I think that
```python3
warnings.filterwarnings("ignore", message=".*findfont.*")
```
I will commit this change and then someone who uses linux can test it. This error shows up on a bunch of cirq notebooks.

tangelo/linq/circuit.py
```diff
@@ -144,6 +146,16 @@ def is_mixed_state(self):
         """
         return "MEASURE" in self.counts
 
+    def draw(self):
+        """Method to output a prettier version of the circuit for use in jupyter notebooks that uses cirq SVGCircuit"""
+        # circular import
+        from tangelo.linq.translator.translate_cirq import translate_c_to_cirq
+        warnings.filterwarnings("ignore", message=".*findfont.*")
```
ValentinS4t1qbit, line 153: This does not seem to work for me (I don't understand your pattern btw, I'm not sure why the dots are here).
Let's consider this a Cirq issue and remove this line as it does not work.

ValentinS4t1qbit (approved):
Thank you !

ValentinS4t1qbit (approved):

PR #268, opened 2023-01-16
Cirq, qulacs, pennylane and projectq operator translation functions

AlexandreF-1qbit:
Highlights:
- Added 2-way translation of qubit operator in several open-source packages (cirq, qulacs, pennylane, projectq). The translation in qulacs can now be handled by the proper function (issue https://github.com/qulacs/qulacs/issues/303 resolved);
- Added testsfor the translation from and to tangelo qubit operators;
- Used the `translate_operator` in the `target_cirq.py` and `target_qulacs.py` simulator files;
- Added docs to `QulacsSimulator.expectation_value_from_prepared_state` method.
- Moved the qubit operator cleaning into the main `translate_operator` function.

EDIT: @JamesB-1qbit asked me a question about the time execution for the qulacs translation. Apparently, the qulacs function was way slower (maybe due to the memory leak). Here are the times for 100 random 50-qubit operators -> 0.09 +- 0.05s (qulacs) vs 0.05 +- 0.03s (custom code with a loop). I think it is better to use external code, as we won't have the task of maintaining it.

### DIFF
--- a/tangelo/linq/translator/translate_cirq.py
+++ b/tangelo/linq/translator/translate_cirq.py
@@ -24,6 +24,7 @@ necessary to account for:
 
 from math import pi
 
+from tangelo.toolboxes.operators import QubitOperator
 from tangelo.helpers import deprecated
 
 
@@ -164,3 +165,52 @@ def translate_c_to_cirq(source_circuit, noise_model=None, save_measurements=Fals
                     target_circuit.append(depo(*depo_list))  # gates targeted
 
     return target_circuit
+
+
+def translate_op_to_cirq(qubit_operator):
+    """Helper function to translate a Tangelo QubitOperator to a cirq linear
+    combination of Pauli strings.
+
+    Args:
+        qubit_operator (tangelo.toolboxes.operators.QubitOperator): Self-explanatory.
+
+    Returns:
+        (cirq.ops.linear_combinations.PauliSum): Cirq summation of Pauli strings.
+    """
+    from openfermion.transforms import qubit_operator_to_pauli_sum
+
+    return qubit_operator_to_pauli_sum(qubit_operator)
+
+
+def translate_op_from_cirq(qubit_operator):
+    """Helper function to translate a cirq linear combination of Pauli strings
+    to a Tangelo QubitOperator.
+
+    Args:
+        qubit_operator (cirq.ops.linear_combinations.PauliSum): Self-explanatory.
+
+    Returns:
+        (tangelo.toolboxes.operators.QubitOperator): Tangelo qubit operator.
+    """
+    import cirq
+
+    # Create a dictionary to append all terms at once.
+    tangelo_op = QubitOperator()
+
+    cirq_pauli_to_string = {cirq.X: "X", cirq.Y: "Y", cirq.Z: "Z"}
+
+    for pauli_word in qubit_operator:
+
+        term_string = ""
+        for line_qubit, cirq_pauli in pauli_word.items():
+
+            term_string += f"{cirq_pauli_to_string[cirq_pauli]}{line_qubit.x} "
+
+        term_string.strip()
+
+        tangelo_op += QubitOperator(term_string.strip(), pauli_word.coefficient)
+
+    # Clean the QubitOperator.
+    tangelo_op.compress()
+
+    return tangelo_op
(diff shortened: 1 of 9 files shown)


tangelo/linq/translator/translate_qubitop.py
```diff
@@ -50,10 +62,15 @@ def translate_operator(qubit_operator, source, target, n_qubits=None):
...
             raise NotImplementedError(f"Qubit operator conversion from {source} to {target} is not supported.")
         qubit_operator = TO_TANGELO[source](qubit_operator)
+        # Clean the very small terms.
+        qubit_operator.compress()
     if target != "tangelo":
         if target not in FROM_TANGELO:
             raise NotImplementedError(f"Qubit operator conversion from {source} to {target} is not supported.")
         n_qubits = count_qubits(qubit_operator) if n_qubits is None else n_qubits
-        qubit_operator = FROM_TANGELO[target](qubit_operator, n_qubits)
+        try:
```
AlexandreF-1qbit, line 71: Would you prefer adding `n_qubits` to all underlying functions instead of putting a try block here?

JamesB-1qbit (approved):

tangelo/linq/translator/translate_qubitop.py
```diff
@@ -50,10 +62,15 @@ def translate_operator(qubit_operator, source, target, n_qubits=None):
         if source not in TO_TANGELO:
             raise NotImplementedError(f"Qubit operator conversion from {source} to {target} is not supported.")
         qubit_operator = TO_TANGELO[source](qubit_operator)
+        # Clean the very small terms.
+        qubit_operator.compress()
```
ValentinS4t1qbit, line 66: How would you feel if, as a user, you pushed data from an existing project into tangelo using a translate function and got something that did not have the same number of terms ?

This function translates operators from a format to another, I do not think we should take the liberty to "clean things up", here.

tangelo/linq/translator/translate_cirq.py
```diff
@@ -164,3 +165,45 @@ def translate_c_to_cirq(source_circuit, noise_model=None, save_measurements=Fals
...
+
+    Args:
+        qubit_operator (cirq.ops.linear_combinations.PauliSum): Self-explanatory.
+
+    Returns:
+        (tangelo.toolboxes.operators.QubitOperator): Tangelo qubit operator.
+    """
+    import cirq
+
+    tangelo_op = QubitOperator()
```
ValentinS4t1qbit, line 197: I think this line can be discarded if L207 reads `=` instead of `+=` ?

tangelo/linq/translator/translate_qubitop.py
ValentinS4t1qbit: I'd rather have a support for `n_qubits` everywhere possible (if we don't know how to enforce it for a specific backend, we can raise a warning if the user provides a value).

Let's make sure `n_qubits` is tagged as optional when it can be.

ValentinS4t1qbit (changes requested):
Good work, some comments here and there but nothing major, I think you got this.

tangelo/linq/translator/translate_qubitop.py
AlexandreF-1qbit: It was previously in each individual function, but I understand the argument of keeping it consistent with the input. I will remove that line.

tangelo/linq/translator/translate_qubitop.py
AlexandreF-1qbit: Alright, I'll try to leverage the `n_qubits` in the function to explicitly set the number of qubits. However I will keep it optional.

tangelo/linq/translator/translate_cirq.py
AlexandreF-1qbit: I think it would work with only one Pauli word, but not with a summation of Pauli words.

tangelo/linq/translator/translate_qubitop.py
```diff
@@ -35,7 +47,9 @@ def translate_operator(qubit_operator, source, target, n_qubits=None):
         qubit_operator (source format): Self-explanatory.
         source (string): Identifier for the source format.
         target (string): Identifier for the target format.
-        n_qubits (int): Number of qubits relevant to the operator.
+        n_qubits (int): Number of qubits relevant to the operator. It is
```
ValentinS4t1qbit, line 50: We need to stop formulating comments that are so specific that they won't be true anymore every time code changes somewhere. We could instead say:

`Number of qubits relevant to the operator. Not used by all format conversion functions, will be computed automatically if not specified.`

ValentinS4t1qbit (approved):

PR #274, opened 2023-02-07
removed examples folder, added submodule Tangelo-Examples

JamesB-1qbit:
Removed examples folder and added link to Tangelo-Examples as submodule.


ValentinS4t1qbit (commented):
Thank you James. I think I only have one question here:

I naively thought we'd just delete that folder and simply update our README / CONTRIBUTION files to redirect the reader to the other repo.

- What is the added value that you see of adding `examples` as a submodule ?
- What is the added cost for us and other users or handling examples and notebooks as a submodule ?

Are we just trading 2 simple `git clone` commands for a single more complicated one (`--recursive` or something) to actually get the contents of that folder and make that sit next to Tangelo code files, for the devs ?

I believe that we'd have to drag and update the commit number of `examples` (eg `Tangelo-Examples`) every time we make changes to it - we had to I think when we updated `agnostic_simulator`, once a dependency of `qsdk`.

If there's no clear reason why we are doing that, I'd rather have the simplest, cleanest solution that avoids tying code that does not need to be tied to anything. Let me know what you think seems to make most sense.

JamesB-1qbit:
I was looking at the insights and a fair number of unique visitors click on examples. I didn't want to lose those people. If we made the link to Tangelo-Examples more prominent in the README, that may be a suitable substitute.
The added cost is occasional updates of changing the commit the folder points to. This involves
1) `cd examples`, 2) `git pull origin main` 3) `git submodule update --remote --merge` 4) `cd ..` 5) `git add examples` 6) `git commit -m "updated to latest examples commit"` 7) `git push`.

The main other benefit is that we can perform testing off of main in Tangelo-Examples and only update the notebooks when a new release is made.

This may be more of an annoyance than is worth it but I was interested in checking it out.

> Thank you James. I think I only have one question here:
> 
> I naively thought we'd just delete that folder and simply update our README / CONTRIBUTION files to redirect the reader to the other repo.
> 
> * What is the added value that you see of adding `examples` as a submodule ?
> * What is the added cost for us and other users or handling examples and notebooks as a submodule ?
> 
> Are we just trading 2 simple `git clone` commands for a single more complicated one (`--recursive` or something) to actually get the contents of that folder and make that sit next to Tangelo code files, for the devs ?
> 
> I believe that we'd have to drag and update the commit number of `examples` (eg `Tangelo-Examples`) every time we make changes to it - we had to I think when we updated `agnostic_simulator`, once a dependency of `qsdk`.
> 
> If there's no clear reason why we are doing that, I'd rather have the simplest, cleanest solution that avoids tying code that does not need to be tied to anything. Let me know what you think seems to make most sense.

ValentinS4t1qbit:
Thank you James.

0. Essentially you're saying that by making `Tangelo-Examples` a submodule of the main repo, the insights of the main repo would also account for visitors who went to `Tangelo-Examples` ?

1. If we can make this a script and drop it in `dev_tools` I think that would be rad (we'll of course have to test it). 
Instruction 6 does not specify which files to commit but I suspect if we add `-a` before `-m` then it seems like it could just be run. Do these instructions assume we are on a particular branch ? Pushing on main would likely not work, per our branch protection policy.

JamesB-1qbit:
0. I think we can track this through insights in Tangelo-Examples as links from github.
1. I looked into this, it seemed more complicated than I would like. It did appear that it was possible to open a PR every time there was a change on Tangelo-Examples but I don't trust it to be dependable.
2. Right now, Tangelo-Examples github action runs tests off of the develop branch. It could have been possible to run the tests off of develop right before a new release and then fixed the examples link to that commit. 

In any case, I removed the examples submodule and simply added a link to Tangelo-Examples higher in index.rst. I think that should be fine for successfully redirecting people.

> Thank you James.
> 
> 0. Essentially you're saying that by making `Tangelo-Examples` a submodule of the main repo, the insights of the main repo would also account for visitors who went to `Tangelo-Examples` ?
> 1. If we can make this a script and drop it in `dev_tools` I think that would be rad (we'll of course have to test it).
>    Instruction 6 does not specify which files to commit but I suspect if we add `-a` before `-m` then it seems like it could just be run. Do these instructions assume we are on a particular branch ? Pushing on main would likely not work, per our branch protection policy.
> 2. I'm not sure I understand. If I want to test notebooks off main, all I need is to have tangelo installed in my environment from the main branch, go into the clone of Tangelo-examples repo and run the notebooks. Likewise, the notebooks can be updated separately after a release.

ValentinS4t1qbit (approved):
Thank you James !

PR #278, opened 2023-02-17
Bugfix: expectation of imaginary qubit_op with desired meas result

JamesB-1qbit:
I found an error in the `get_expectation_value` with a `desired_meas_result` and non-real `QubitOperator`


ValentinS4t1qbit (approved):
Thank you for thinking about this !

PR #285, opened 2023-02-25
Depth function performance increase

ValentinS4t1qbit:
Just a look up dictionary really. No increase in memory, we apparently go faster than cirq itself to compute depth from scratch on circuits of depth 100K+ and 150K gates. Things are instant.

### DIFF
--- a/tangelo/linq/circuit.py
+++ b/tangelo/linq/circuit.py
@@ -195,30 +195,44 @@ class Circuit:
 
     def depth(self):
         """ Return the depth of the quantum circuit, by computing the number of moments. Does not count
-        qubit initialization as a moment (unlike Cirq, for example).
+        qubit initialization as a moment (unlike Cirq, for example). Computes from scratch.
         """
+        # List of qubit indices involved in each moment. Look up dict for latest moment for each index.
         moments = list()
+        latest_moment = dict()
 
-        for g in self._gates:
+        # Traverse gates and compute moments
+        for g in self:
             qubits = set(g.target) if g.control is None else set(g.target + g.control)
 
             if not moments:
                 moments.append(qubits)
+                for i in qubits:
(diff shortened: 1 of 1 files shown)


tangelo/linq/circuit.py
```diff
@@ -195,31 +195,45 @@ def check_index_valid(index):
...
+                # Case 1: Gate can be included in a previous moment
+                # Includes b = -1 case where all qubits are encountered for the 1st time
+                if (b + 1) < len(moments):
+                    moments[b + 1] = moments[b + 1] | qubits
+                # Case 2: Gate is part of a new moment
+                else:
+                    moments.append(qubits)
         return len(moments)
 
+        # Option 2: No moment info. Thats the whole thing. 2x speedup on h20_321g example
```
ValentinS4t1qbit, line 228: I left option 2 around, see comment in issue open by @AlexandreF-1qbit .
We just need to quickly decide which one we pick and I clean up.

AlexandreF-1qbit: I also prefer the 1st option, as it seems to be reasonable on the memory side, and it could be useful info for "manual" analysis if we want to get this information (now it is ok to use the `len` and throw it away).

tangelo/linq/circuit.py
```diff
@@ -195,31 +195,45 @@ def check_index_valid(index):
 
     def depth(self):
         """ Return the depth of the quantum circuit, by computing the number of moments. Does not count
-        qubit initialization as a moment (unlike Cirq, for example).
+        qubit initialization as a moment (unlike Cirq, for example). Computes from scratch.
```
AlexandreF-1qbit, line 198: Compute?

AlexandreF-1qbit (approved):

JamesB-1qbit (approved):

PR #289, opened 2023-03-08
Translation perf tests (operator, circuit)

ValentinS4t1qbit:


tangelo/linq/tests/test_translator_perf.py
```diff
@@ -0,0 +1,91 @@
...
+
+class PerfTranslatorTest(unittest.TestCase):
+
+    def test_perf_operator(self):
+        """ Performance test with a reasonable large input for operator """
+
+        print(f'\n[Performance Test :: linq operator format conversion]')
+        print(f'\tInput size: n_qubits={n_qubits_op}, n_terms={n_terms}\n')
+
+        for f in FROM_TANGELO_OP:
```
ValentinS4t1qbit, line 55: Right now:

- try / except blocks are here because we don't care about errors (we already have numerical tests somewhere else) and to ignore any issue arising from backends not being installed. e.g it will try to run everything it finds
- currently only try to translate to tangelo for whatever is supported the other way.

For operator, this covers everything. For circuits, this does not cover the very few formats whose translation is only supported to tangelo, but it does not seem like anything users have cared about so far (also, supporting them means creating the input circuit in their format and either reading from file or a variable) and we know we do not use backends relying on these formats yet.

AlexandreF-1qbit (approved):

PR #291, opened 2023-03-09
Re-enabling open-shell DMET

AlexandreF-1qbit:
Highlights
- Re-enable UHF+DMET.
- One thing was missing to fully support the DMET + UHF, i.e. the padding of frozen orbitals. A new function has been added (with a relevant test).


tangelo/toolboxes/molecular_computation/rdms.py
```diff
@@ -292,3 +292,155 @@ def pad_rdms_with_frozen_orbitals(sec_mol, onerdm, twordm):
...
+
+    for i, j in it.product(range(n_occ0_a), repeat=2):
+        twordm_aa[i, i, j, j] -= 1
+        twordm_aa[i, j, j, i] += 1
+
+    for i, j in it.product(range(n_occ0_b), repeat=2):
+        twordm_bb[i, i, j, j] -= 1
+        twordm_bb[i, j, j, i] += 1
+
+    for i in range(n_occ0_a):
```
ValentinS4t1qbit, line 381: you imported itertools already, just use it here too.

tangelo/toolboxes/molecular_computation/tests/test_rdms.py
```diff
@@ -131,22 +133,50 @@ def test_compute_rdms_from_classical_shadow(self):
...
 
         np.testing.assert_array_almost_equal(padded_onerdm, test_onerdm, decimal=3)
         np.testing.assert_array_almost_equal(padded_twordm.reshape((7,)*4), test_twordm, decimal=3)
 
+    def test_pad_unrestricted_rdms_with_frozen_orbitals(self):
+        """Test padding of RDMs with frozen orbitals indices (unrestricted)."""
+
+        mol = SecondQuantizedMolecule(mol_H2O_sto3g.xyz, uhf=True, frozen_orbitals=[(0, 3, 4, 5), (0, 3, 4, 5)])
+
+        onerdm_a_to_pad = np.loadtxt(pwd_this_test + "/data/H2O_UHF_sto3g_onerdm_frozen0345_alpha.data")
```
ValentinS4t1qbit, line 157: I am usually not fond of having entire matrices loaded and compared terms by terms in tests.
That usually makes the tests pretty sensitive to changes in the compute package that uses them, stores a large quantity of data (mostly zeros in your case).

If you think this is appropriate and the precision of your assert statement is good enough, then it's alright.
The alternative is to instead make assert statements on properties of these data-structures (their trace? norm? determinant ? sparse structure ? the value of an expression that uses them ?) that ideally has some physical / practical meaning. Right now, your test will simply tell you whether entry by entry, you are close enough to what PYSCF or whatever reference tool you used computed at a given time. If not, well all you know is that the entries aren't the same (but maybe its good, and these guys fixed a bug or something).

For example, the 1 and 2-rdm tests with DMET revealed that computing an expression using the trace of the matrix should match the number of electrons of the system. What's great about this is that regardless of the package we use and the sensitivity of the calculations / bugs in there, is that it focuses on values that truly matter instead of arbitrary quantities that may be meaningless to the reader.

It is up to you and your creativity here.

tangelo/toolboxes/molecular_computation/rdms.py
```diff
@@ -292,3 +292,155 @@ def pad_rdms_with_frozen_orbitals(sec_mol, onerdm, twordm):
...
+               twordm_aa.reshape(n_mos0_a**2, n_mos0_a**2), idx_a, idx_a)
+    takebak_2d(dm2_bb.reshape(n_mos_b**2, n_mos_b**2),
+               twordm_bb.reshape(n_mos0_b**2, n_mos0_b**2), idx_b, idx_b)
+    takebak_2d(dm2_ab.reshape(n_mos_a**2, n_mos_b**2),
+               twordm_ab.reshape(n_mos0_a**2, n_mos0_b**2), idx_a, idx_b)
+    twordm_aa_padded, twordm_ab_padded, twordm_bb_padded = dm2_aa, dm2_ab, dm2_bb
+
+    # The next few lines will reembed the on rdm, but with the frozen orbital
+    # elements of the one rdm matrix.
+    onerdm_a_padded_without_diag = np.copy(onerdm_a_padded)
```
ValentinS4t1qbit, line 407: Think we can have a quick function defined in this function that avoids us repeating the same 30 lines of code ?
It looks like you did the same with the non-padded matrices already.

ValentinS4t1qbit (changes requested):
Thank you, just a few comments here.

tangelo/toolboxes/molecular_computation/tests/test_rdms.py
AlexandreF-1qbit: The whole purpose of this function is to pads the RDMs with (2s/1s (depending on UHF or RHF) or 0s). I designed the test so the padding is not trivial (not only the very fist or very last orbitals), so that the padding happens "inside" the numpy arrays.

Evaluating global quantities (trace, norm, determinant) in this situation does not make sense in my opinion, because a misplaced element in the RDMs can have a big impact on the computed eigenvalue (especially with DMET where it is used). Trace could be used to check if the number of electrons increases, but it would not tell if the off-diagonal are OKs.

tangelo/toolboxes/molecular_computation/rdms.py
AlexandreF-1qbit: I don't a way to be more concise at the moment. The goal is to "unpack" the RDM diagonals, reinstate missing columns and "repack" them at the end. This operation is done twice, but the first time in reverse (+ -> -, - -> +, etc.)

I don't think we would need those function elsewhere in a typical workflow. I also think it would impede readability to introduce (at least) 2 more functions to take into account the AA(BB) + AB terms in the 2-RDMs. In short, it requires some thoughts.

tangelo/toolboxes/molecular_computation/tests/test_rdms.py
ValentinS4t1qbit: Alright then, we'll keep it that way :)

Another way to test this would be to just track the positions / number of zero and non-zero elements, and maybe check that the sum of all elements in the matrix does not change before and after padding (I'm not saying this is what we should do, just giving you an example of how creative we can be). Happy to proceed with what you have :)

tangelo/toolboxes/molecular_computation/rdms.py
ValentinS4t1qbit: Then it does not sound worth our time atm :)
Happy to leave it as that

ValentinS4t1qbit (approved):
Thank you ! When tests pass, I see no reason not to merge this.

The last important step for OS DMET would be to have some notebook content. We don't know if people will push this project for us, so I think we should put enough of our findings out there to just generate enough excitement. Maybe this can also be a blog article. If that makes some research group somewhere use our code and even publish results, that would be rad.

PR #300, opened 2023-05-05
Supporting symbols as parameters and add nsimplify to results

AlexandreF-1qbit:
Highlights:
- The best workflow I found to simplify the sympy expressions is to do a `simplify` then a `evalf` on the results.
- Add the `and n_qubits is None` condition to a check in `translate_qubitop`.
- Removed small probabilities from the bitstring dict.

EDIT: Removed `nsimplify` -> switched to `evalf()`.

### DIFF
--- a/tangelo/linq/gate.py
+++ b/tangelo/linq/gate.py
@@ -20,6 +20,8 @@ from math import pi
 from typing import Union
 
 from numpy import integer, ndarray, floating
+from sympy import Symbol
+
 
 ONE_QUBIT_GATES = {"H", "X", "Y", "Z", "S", "T", "RX", "RY", "RZ", "PHASE"}
 TWO_QUBIT_GATES = {"CNOT", "CX", "CY", "CZ", "CRX", "CRY", "CRZ", "CPHASE", "XX", "SWAP"}
@@ -172,7 +174,7 @@ class Gate(dict):
 
         if self.parameter == "":
             new_parameter = ""
-        elif isinstance(self.parameter, (float, floating, int, integer)):
+        elif isinstance(self.parameter, (float, floating, int, integer, Symbol)):
             new_parameter = -self.parameter
         else:
             raise AttributeError(f"{self.name} is not an invertible gate when parameter is {self.parameter}")
--- a/tangelo/linq/target/target_sympy.py
+++ b/tangelo/linq/target/target_sympy.py
@@ -99,13 +99,14 @@ class SympySimulator(Backend):
             sympy.core.add.Add: Eigenvalue represented as a symbolic sum.
         """
 
-        from sympy import simplify, nsimplify
-        from sympy.physics.quantum import qapply, Dagger
+        from sympy import simplify, nsimplify, cos
+        from sympy.physics.quantum import Dagger
 
         prepared_state = self._current_state if prepared_state is None else prepared_state
         operator = translate_operator(qubit_operator, source="tangelo", target="sympy", n_qubits=n_qubits)
 
         eigenvalue = Dagger(prepared_state) * operator * prepared_state
+        eigenvalue = eigenvalue[0, 0].rewrite(cos)
         eigenvalue = simplify(eigenvalue)
         eigenvalue = nsimplify(eigenvalue, tolerance=1e-4, rational=True)
 
--- a/tangelo/linq/target/target_sympy.py
+++ b/tangelo/linq/target/target_sympy.py
@@ -48,7 +48,7 @@ class SympySimulator(Backend):
                 by the user (if not, set to None).
         """
 
-        from sympy import simplify
+        from sympy import simplify, nsimplify
         from sympy.physics.quantum import qapply
         from sympy.physics.quantum.qubit import Qubit, matrix_to_qubit, \
             qubit_to_matrix, measure_all
@@ -75,8 +75,10 @@ class SympySimulator(Backend):
         frequencies = dict()
         for vec, prob in measurements:
             prob = simplify(prob, tolerance=1e-4)
-            bistring = "".join(str(bit) for bit in reversed(vec.qubit_values))
-            frequencies[bistring] = prob
+            prob = nsimplify(prob, tolerance=1e-4, rational=True)
+            if not prob.is_zero:
+                bistring = "".join(str(bit) for bit in reversed(vec.qubit_values))
+                frequencies[bistring] = prob
 
         return (frequencies, python_statevector) if return_statevector else (frequencies, None)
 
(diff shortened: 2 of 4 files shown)


tangelo/linq/gate.py
```diff
@@ -20,6 +20,8 @@
 from typing import Union
 
 from numpy import integer, ndarray, floating
+from sympy import Symbol
```
ValentinS4t1qbit, line 23: with this, `sympy` is now a main dependency of Tangelo. I just want us to acknowledge that: users can no longer use Tangelo with having sympy in their environment after this change is integrated.

tangelo/linq/target/target_sympy.py
```diff
@@ -75,8 +75,10 @@ def simulate_circuit(self, source_circuit: Circuit, return_statevector=False, in
         frequencies = dict()
         for vec, prob in measurements:
             prob = simplify(prob, tolerance=1e-4)
-            bistring = "".join(str(bit) for bit in reversed(vec.qubit_values))
-            frequencies[bistring] = prob
+            prob = nsimplify(prob, tolerance=1e-4, rational=True)
```
ValentinS4t1qbit, line 78: This line is what enables us to have rational approximations of floats up to a given precision, right ?
Out of curiosity: do you know how time-consuming is it and how helpful do you think this is in practice ?

I was naively expecting fractions to be exact in the output expressions.

(commenting this line on 1-2 use cases of decent size will tell you how much time it takes)

tangelo/linq/target/target_sympy.py
```diff
@@ -97,14 +99,18 @@ def expectation_value_from_prepared_state(self, qubit_operator, n_qubits, prepar
...
 
         prepared_state = self._current_state if prepared_state is None else prepared_state
         operator = translate_operator(qubit_operator, source="tangelo", target="sympy", n_qubits=n_qubits)
-        eigenvalue = qapply(Dagger(prepared_state) * operator * prepared_state)
 
-        return simplify(eigenvalue)
+        eigenvalue = Dagger(prepared_state) * operator * prepared_state
+        eigenvalue = eigenvalue[0, 0].rewrite(cos)
+        eigenvalue = simplify(eigenvalue)
+        eigenvalue = nsimplify(eigenvalue, tolerance=1e-4, rational=True)
```
ValentinS4t1qbit, line 111: see comment above

tangelo/linq/translator/translate_qubitop.py
```diff
@@ -76,7 +76,7 @@ def translate_operator(qubit_operator, source, target, n_qubits=None):
         if target in {"qiskit", "sympy"}:
             # The count_qubits function has no way to detect the number of
             # qubits when an operator is only a tensor product of I.
-            if qubit_operator == QubitOperator((), qubit_operator.constant):
+            if qubit_operator == QubitOperator((), qubit_operator.constant) and n_qubits is None:
```
ValentinS4t1qbit, line 79: This does not flow well with L81, that says " if n_qubits is None, then its value is count_qubits(q_op)".

Either that conditional statement at L81 is not necessary anymore (since it will never be encountered as L79 prevents it) or L79 is not the solution.

We need the case for both sympy and qiskit to work, make sure the behaviour is as expected for them (with or without n_qubits).

tangelo/linq/translator/translate_sympy.py
```diff
@@ -212,15 +212,19 @@ def translate_op_to_sympy(qubit_operator, n_qubits):
...
             objects.
     """
-    from sympy import Identity
-    from sympy.physics.paulialgebra import Pauli
+    from sympy import Matrix, I, zeros
     from sympy.physics.quantum import TensorProduct
 
     # Pauli string to sympy Pauli algebra objects.
-    map_to_paulis = {"I": Identity(1), "X": Pauli(1), "Y": Pauli(2), "Z": Pauli(3)}
+    map_to_paulis = {
```
ValentinS4t1qbit, line 219: What were the consequences of this changes, out of curiosity ?

ValentinS4t1qbit (changes requested):
A few small things :)

tangelo/linq/gate.py
AlexandreF-1qbit: Openfermion is already requiring explicitly `sympy` (https://github.com/quantumlib/OpenFermion/blob/master/requirements.txt). However, we could put checkers, but in this case, I don't think it is necessary,

tangelo/linq/target/target_sympy.py
AlexandreF-1qbit: It is, but I found out it is not simpler to leave fractions: in my use-cases, I was getting exact fractions in the form of `3290239049023/348923890208423094823` or something uglier like that. I prefer getting `9.42967547e-9` than the fraction in this case.

In the case of speed, it will need investigation. This is also why I set a loose tolerance, so it does not look for a machine-precision exact fraction.

tangelo/linq/translator/translate_sympy.py
AlexandreF-1qbit: Instead of getting 
```
sin(theta/2)**2*<11|*sigma2xsigma2xsigma1xsigma1*|11> + ...
```
for an expectation value, it is applying the operations (<11|11>=1, and so on, and we get an imaginary number as an expectation value).

tangelo/linq/translator/translate_qubitop.py
AlexandreF-1qbit: This code raises `ValueError: The number of qubits (n_qubits) must be provided.`, even if I provided `n_qubits`. If we don't check for `n_qubits`, it will fail for every empty `QubitOperator`. The `count_qubits` can automatically get the number of qubits of an operator, but it fails with an empty `QubitOperator`.

```python
from tangelo.linq import translate_operator

op = QubitOperator((), 5.)
translate_operator(op, source="tangelo", target="qiskit", n_qubits=5)
```

tangelo/linq/gate.py
ValentinS4t1qbit: Thank you, I had a feeling it was already a dependency :)
I think there's nothing to do here, we're good

tangelo/linq/target/target_sympy.py
ValentinS4t1qbit: Sorry, what I meant is, is there value in request a fraction that is an approximation of a floating point number, rather than requesting the floats, up to a level of precision ?
(e.g `0.71428` instead of `5/7` etc when we know 5/7 is actually not exact and can't be directly compared to machine calculations, stored in floats)

(for small numbers, such as your example, scientific notation is nice unless we proclaim things below a given threshold to be 0)

tangelo/linq/translator/translate_qubitop.py
ValentinS4t1qbit: Thank you. If that means all use cases are covered and handled as they should, then you can resolve this conversation :)

tangelo/linq/translator/translate_sympy.py
ValentinS4t1qbit: Great, thanks.

ValentinS4t1qbit (approved):
Good work !

We note that having the cutoff threshold for `simplify` as a backend option may be an improvement in the future.

PR #302, opened 2023-05-17
Trim trivial qubits from circuit and Hamiltonian

elloyd-1qbit:
Given a circuit and Hamiltonian these functions will:

-  Calculate expectation values on qubits which are unentangled and in the |0> or |1> state.
- Incorporate these expectation values (either 0, +/-1) into the Hamiltonian, potentially reducing the number of terms.


ValentinS4t1qbit (approved):
Thank you Erika !

PR #304, opened 2023-05-23
Set the DMET virtual orbital truncation threshold at the user level

AlexandreF-1qbit:
Highlights:
- Make a `virtual_orbital_threshold` available. Before, it was hard-coded at `1e-13`. Changing this value gives us the ability of trading virtual orbital truncation and DMET energy accuracy. This is especially useful when working with small basis sets.
- Add a verbose print to output the occupancy eigenvalues (those are compared to the threshold).
- Force the occupancy eigenvalues to their absolute value (may be negative due to numerical noise, and this was introducing inconsistencies).


ValentinS4t1qbit (approved):
I'm good with the changes, thank you

PR #307, opened 2023-05-31
added typing for option dictionaries

JamesB-1qbit:
Added typing for option dictionaries. Most attributes should now be available to view in a persons coding environment. For example, in VSCode.
Screen Shot 2023-05-31 at 10 58 46 AM

### DIFF
--- a/tangelo/algorithms/variational/iqcc_ilc_solver.py
+++ b/tangelo/algorithms/variational/iqcc_ilc_solver.py
@@ -32,11 +32,11 @@ Refs:
 from typing import List, Union, Optional
 
 from tangelo import SecondQuantizedMolecule
-from tangelo.toolboxes.operators import QubitOperator
+from tangelo.algorithms.variational.vqe_solver import VQESolver
 from tangelo.linq import get_backend
+from tangelo.toolboxes.operators import QubitOperator
 from tangelo.toolboxes.ansatz_generator.ilc import ILC
 from tangelo.toolboxes.ansatz_generator.qcc import QCC
-from tangelo.algorithms.variational.vqe_solver import VQESolver
 from tangelo.toolboxes.ansatz_generator._qubit_ilc import ilc_op_dress
 
 
--- a/tangelo/algorithms/variational/iqcc_ilc_solver.py
+++ b/tangelo/algorithms/variational/iqcc_ilc_solver.py
@@ -29,7 +29,10 @@ Refs:
     2. R. A. Lang, I. G. Ryabinkin, and A. F. Izmaylov.
         J. Chem. Theory Comput. 2021, 17, 1, 66–78.
 """
+from typing import List, Union, Optional
 
+from tangelo import SecondQuantizedMolecule
+from tangelo.toolboxes.operators import QubitOperator
 from tangelo.linq import get_backend
 from tangelo.toolboxes.ansatz_generator.ilc import ILC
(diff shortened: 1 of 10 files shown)


tangelo/algorithms/variational/iqcc_ilc_solver.py
```diff
@@ -29,7 +29,10 @@
     2. R. A. Lang, I. G. Ryabinkin, and A. F. Izmaylov.
         J. Chem. Theory Comput. 2021, 17, 1, 66–78.
 """
+from typing import List, Union, Optional
 
+from tangelo import SecondQuantizedMolecule
```
ValentinS4t1qbit, line 34: I think we are importing those two just to enable typing. I'm trying to understand if there are unwanted consequences to these imports.

It seems perfectly acceptable for our algorithm to depend of these data-types, and its execution would require them anyway I believe (we eventually will have a QubitOperator, and as for SecondQuantizedMolecule it should hopefully not require pyscf/psi4 if the user is actually not throwing in a molecule but directly a qubit Hamiltonian at the algorithm) . Is that correct ?

If yes you can resolve0

tangelo/algorithms/variational/iqcc_ilc_solver.py
```diff
@@ -106,6 +109,25 @@ def __init__(self, opt_dict):
...
+
+        self.molecule: SecondQuantizedMolecule = copt_dict.pop("molecule", None)
+        self.qubit_mapping: str = copt_dict.pop("qubit_mapping", "jw")
+        self.up_then_down: str = copt_dict.pop("up_then_down", False)
+        self.initial_var_params: Union[List, str] = copt_dict.pop("initial_var_params", None)
+        self.backend_options: dict = copt_dict.pop("backend_options", default_backend_options)
+        self.penalty_terms: Optional[dict] = copt_dict.pop("penalty_terms", None)
+        self.ilc_ansatz_options: dict = copt_dict.pop("ilc_ansatz_options", dict())
+        self.qcc_ansatz_options: dict = copt_dict.pop("qcc_ansatz_options", dict())
+        self.qubit_hamiltonain: QubitOperator = copt_dict.pop("qubit_hamiltonian", None)
```
ValentinS4t1qbit, line 122: self.qubit_hamilton**ian** (you have a typo).

I'm a bit scared here because the rest of the code relies on `self.qubit_hamiltonian` and no error in our tests was caught.

tangelo/algorithms/variational/iqcc_ilc_solver.py
```diff
@@ -106,6 +109,25 @@ def __init__(self, opt_dict):
             else:
                 raise KeyError(f"The keyword {param} is not available in self.__class__.__name__.")
 
+        copt_dict = opt_dict.copy()
```
ValentinS4t1qbit, line 112: In this file I think you forgot to remove the previous code block, which is now replaced by your solution (just keep default_backend_options)

ValentinS4t1qbit (changes requested):
Small things. You may want to add a test in iqcc_ilc_solver (doesnt hurt)

tangelo/algorithms/variational/iqcc_ilc_solver.py
JamesB-1qbit: The error was caught in testing when I removed the old assignment from dictionary method.

ValentinS4t1qbit (approved):
Looks good to me, thank you James.

PR #308, opened 2023-06-01
Change multi-controlled CNOT to multi-controlled CX

JamesB-1qbit:
Previously CNOT ignored all controls except the first without warning. This changes the name in the circuit to CX.

### DIFF
--- a/tangelo/linq/tests/test_translator_circuit.py
+++ b/tangelo/linq/tests/test_translator_circuit.py
@@ -19,6 +19,7 @@
 
 import unittest
 import os
+
 import numpy as np
 
 from tangelo.linq import Gate, Circuit
@@ -67,11 +68,10 @@ class TranslateCircuitTest(unittest.TestCase):
             VS one obtained through translation from an abstract format
         """
         import qulacs
-
-        # Generates the qulacs circuit by translating from the abstract one
+        
+        # Test 1: Simulation of trasnlated circuit VS native circuit
+        # Generates the qulacs circuit by translating from the abstract one, run the simulation afterwards
         translated_circuit = translate_c(abs_circ, "qulacs")
-
-        # Run the simulation
         state1 = qulacs.QuantumState(abs_circ.width)
         translated_circuit.update_quantum_state(state1)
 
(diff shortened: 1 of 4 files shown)


tangelo/linq/tests/test_translator_circuit.py
```diff
@@ -32,7 +32,9 @@
 gates = [Gate("H", 2), Gate("CNOT", 1, control=0), Gate("CNOT", 2, control=1), Gate("Y", 0), Gate("S", 0)]
 abs_circ = Circuit(gates) + Circuit([Gate("RX", 1, parameter=2.)])
 multi_controlled_gates = [Gate("X", 0), Gate("X", 1), Gate("CX", target=2, control=[0, 1])]
+multi_controlled_gates2 = [Gate("X", 0), Gate("X", 1), Gate("CNOT", target=2, control=[0, 1])]
 abs_multi_circ = Circuit(multi_controlled_gates)
+abs_multi_circ2 = Circuit(multi_controlled_gates)
```
ValentinS4t1qbit, line 37: I think you wanted `multi_controlled_gates2` here.

tangelo/linq/tests/test_translator_circuit.py
```diff
@@ -236,6 +247,11 @@ def test_cirq(self):
 
         np.testing.assert_array_equal(v1, v2)
 
+        job_sim = cirq_simulator.simulate(translated_circuit2)
```
ValentinS4t1qbit, line 250: These tests have become confusing overall. One test function is now a bunch of copy pastes that blend together. We don't need to repeat the same comments every time, we should now introduce comments that say what the next 10 lines are doing etc to clearly delineate the "sub tests" that you are running here.

qulacs just repeats the same blocks and comments on and on, with no clear structure. cirq has no comments at all.

ValentinS4t1qbit (changes requested):
A few things to clean up in the tests, which are now a bit confusing.

We are testing a few different scenarios but they are not clear to the reader. It's ok if the first testin each method is detailed but the ones right after can just be introduced by 1-2 lines of comment (what scenario they test) and then we can just group most code lines together: the reader knows how these work by then.

tangelo/linq/tests/test_translator_circuit.py
JamesB-1qbit: Our translation of qiskit does not need to be changed. It currently does not support more than single controls at this point. Previously, I believe that qiskit just had multi-controlled X gates but they have a few others now. Maybe something to look into.

ValentinS4t1qbit: Thank you. We do not need to look into this actively, there's no clear demand

ValentinS4t1qbit (approved):

PR #309, opened 2023-06-13
FCISolverPsi4

JamesB-1qbit:
Changes `FCISolver` to calling function, adds `get_fci_solver`.
Now `FCISolverPsi4` and `FCISolverPySCF`.


tangelo/algorithms/classical/fci_solver.py
```diff
@@ -15,12 +15,25 @@
...
+from typing import Union, Type
+import warnings
 
+import numpy as np
+
+from tangelo.toolboxes.molecular_computation.molecule import SecondQuantizedMolecule
+from tangelo.toolboxes.molecular_computation import IntegralSolverPsi4, IntegralSolverPySCF
 from tangelo.algorithms.electronic_structure_solver import ElectronicStructureSolver
-from tangelo.helpers.utils import is_package_installed
+from tangelo.helpers.utils import installed_chem_backends, deprecated, is_package_installed
```
AlexandreF-1qbit, line 26: deprecated is not needed at the moment if we don't uncomment the decorator.

tangelo/algorithms/classical/fci_solver.py
```diff
@@ -131,3 +144,170 @@ def get_rdm(self):
...
+
+        import psi4
+        self.backend = psi4
+        self.backend.core.clean_options()
+        self.ciwfn = None
+        self.any_close = False
+        if isinstance(molecule.solver, IntegralSolverPsi4) and not molecule.symmetry:
+            self.molecule = molecule
+        else:
+            print(molecule.mo_energies)
```
AlexandreF-1qbit, line 177: Debug print?

tangelo/algorithms/classical/fci_solver.py
```diff
@@ -131,3 +144,170 @@ def get_rdm(self):
...
+
+            for i, val in enumerate(molecule.mo_energies[:-1]):
```
AlexandreF-1qbit, line 179: Could be done without a python loop by rounding the floats and getting the unique elements, then compare the array shapes.
```python
unique_arr = np.unique(arr.round(decimals=tolerance))
self.any_close = (unique_arr.shape != arr.shape)
```
(not tested, at your own risk) Another option would be to look at the count array returned by the `np.unique` function.

AlexandreF-1qbit (changes requested):

tangelo/algorithms/classical/fci_solver.py
```diff
@@ -131,3 +144,170 @@ def get_rdm(self):
...
+
+        Returns:
+            numpy.array: One-particle RDM.
+            numpy.array: Two-particle RDM.
+
+        Raises:
+            RuntimeError: If method "simulate" hasn't been run.
+        """
+
+        # Check if Full CI is performed
```
ValentinS4t1qbit, line 230: has been performed

tangelo/algorithms/classical/fci_solver.py
```diff
@@ -131,3 +144,170 @@ def get_rdm(self):
...
+        if self.ciwfn is None:
+            raise RuntimeError("FCISolver: Cannot retrieve RDM. Please run the 'simulate' method first")
+        if self.any_close:
+            warnings.warn("Degenerate orbitals are present in the molecule. The fci calculation is performed "
```
ValentinS4t1qbit, line 234: `self.degenerate_orbitals` or `self_degenerate` may be a better choice than `self.is_close` then :)

tangelo/algorithms/classical/fci_solver.py
```diff
@@ -131,3 +144,170 @@ def get_rdm(self):
...
+
+
+fci_solver_dict = {"pyscf": FCISolverPySCF, 'psi4': FCISolverPsi4}
+
+
+def get_fci_solver(molecule: SecondQuantizedMolecule, solver: Union[None, str, Type[ElectronicStructureSolver]] = default_fci_solver, **kwargs):
+    """Return requested target backend object.
+
+    Args:
+        solver (string or Type[ElectronicStructureSolver] or None): Supported string identifiers can be found in
```
ValentinS4t1qbit, line 250: missing "molecule" as arg

tangelo/algorithms/classical/fci_solver.py
```diff
@@ -131,3 +144,170 @@ def get_rdm(self):
...
+            fci_solver_dict (from tangelo.algorithms.classical.fci_solver). Can also provide a user-defined backend (child to ElectronicStructureSolver class)
+        kwargs: Other arguments that could be passed to a target. Examples are qubits_to_use for a QPU, transpiler
```
ValentinS4t1qbit, line 252: This does not sound like an appropriate list of examples here for kwargs. What did you envision for them, in this particular context ?

tangelo/algorithms/classical/fci_solver.py
```diff
@@ -131,3 +144,170 @@ def get_rdm(self):
...
+
+    if solver is None:
+        if isinstance(molecule.solver, IntegralSolverPySCF):
+            solver = FCISolverPySCF
+        elif isinstance(molecule.solver, IntegralSolverPsi4):
+            solver = FCISolverPsi4
+        elif default_fci_solver is not None:
+            solver = default_fci_solver
+        else:
+            raise ModuleNotFoundError(f"One of the backends for {list(fci_solver_dict.keys())} needs to be installed to use FCISolver in Tangelo"
```
ValentinS4t1qbit, line 264: remove "in Tangelo". We know this is Tangelo.

tangelo/algorithms/classical/fci_solver.py
```diff
@@ -131,3 +144,170 @@ def get_rdm(self):
...
+                                      "without providing own implementation.")
```
ValentinS4t1qbit, line 265: without providing an implementation ? A user-defined implementation ?

tangelo/algorithms/classical/fci_solver.py
```diff
@@ -131,3 +144,170 @@ def get_rdm(self):
...
+    # If target is a string use target_dict to return built-in backend
```
ValentinS4t1qbit, line 266: skip a line before this comment

tangelo/algorithms/classical/fci_solver.py
```diff
@@ -131,3 +144,170 @@ def get_rdm(self):
...
+    elif isinstance(solver, str):
+        try:
+            solver = fci_solver_dict[solver.lower()]
+        except KeyError:
+            raise ValueError(f"Error: backend {solver} not supported. Available built-in options: {list(fci_solver_dict.keys())}")
+    elif not issubclass(solver, ElectronicStructureSolver):
+        raise TypeError(f"Target must be a str or a subclass of Backend but received class {type(solver).__name__}")
```
ValentinS4t1qbit, line 273: change Backend to the name of the actual base class, here.

tangelo/algorithms/classical/fci_solver.py
```diff
@@ -131,3 +144,170 @@ def get_rdm(self):
...
+
+    return solver(molecule, **kwargs)
+
+
+# @deprecated("Please use get_fci_solver.")
```
ValentinS4t1qbit, line 278: Any reason we should use `get_fci_solver` over `FCISolver` ? It seems to be just a renaming to adhere to PEP8.

We did that for the quantum simulation backends because so many choices (and therefore Backend class names) and it makes sense for developers and users to loop over them or easily switch. I do not remember if there was another reason.

But in this context, what do you dislike about people writing `fci_solver = FCISolver(mol, "pyscf")` ?

tangelo/algorithms/classical/fci_solver.py
```diff
@@ -131,3 +144,170 @@ def get_rdm(self):
...
+def FCISolver(molecule: SecondQuantizedMolecule, solver: Union[None, str, Type[ElectronicStructureSolver]] = default_fci_solver, **kwargs):
+    return get_fci_solver(molecule, solver, **kwargs)
```
ValentinS4t1qbit, line 280: We need a docstring for L279 with the purpose of this.

tangelo/algorithms/classical/fci_solver.py
```diff
@@ -131,3 +144,170 @@ def get_rdm(self):
...
+
+def getswaps(a):
+    """Takes a list and returns the swaps necessary to obtain the ordered version.
+
+    Example: getswaps([4, 3, 2, 1]) -> [[0, 3], [1, 2]]
+    swap [0, 3] results in [1, 3, 2, 4]
+    swap [1, 2] results in [1, 2, 3, 4]
+
+    Args:
+        a (list): Unordered list
```
ValentinS4t1qbit, line 291: Of what types of elements ? Integers ?

tangelo/algorithms/classical/fci_solver.py
```diff
@@ -131,3 +144,170 @@ def get_rdm(self):
...
+
+    Returns:
+        List[list]: The swap operations required to order the list.
```
ValentinS4t1qbit, line 294: It would make more sense to return a list of 2-tuples. Each of this 2-tuple is a swap, an immutable unit of information. 

A list is usually for a mutable data-structure that can grow / shrink, where elements can be processed in any order to accomplish something. It is much more general and vague.

tangelo/algorithms/classical/fci_solver.py
```diff
@@ -131,3 +144,170 @@ def get_rdm(self):
...
+    """
+
+    n = len(a)
```
ValentinS4t1qbit, line 297: Here's my implementation, using np.argsort (or at least an example). It should be much faster on larger lists and is more compact (I think there's 6 lines of code including an import).

When solving what seems to be standard problems, I recommend you try to think immediately about what standard functions and libraries can help in the future.

```
l = [4, 3, 2, 1]

# Use numpy here to find the permutation needed to sort your list.
import numpy as np
arr = np.array(l)
p = np.argsort(l)
print(p)

# Compare to the identity to find your swaps
id = np.arange(0, len(l))
swaps = {(min(id[i], p[i]), max(id[i], p[i])) for i in range(len(l))}
print(swaps)

# If you -really- need an ordered set, for some reason, return the sorted list.
# underlying sort function for tuples sort using first element, which I think is what you want.
swaps_ordered = sorted(swaps)
print(swaps_ordered)

File: tangelo/toolboxes/molecular_computation/integral_solver_psi4.py [hunk 17]
```diff
@@ -85,6 +91,13 @@ def compute_mean_field(self, sqmol):
 
 
         """
+        if sqmol.symmetry:
+            input_string = f"{sqmol.q} {sqmol.spin + 1} \n"
+            for line in sqmol.xyz:
+                input_string += f"{line[0]} {line[1][0]} {line[1][1]} {line[1][2]} \n"
+            if isinstance(sqmol.symmetry, str):
+                input_string += "symmetry"+sqmol.symmetry
```
ValentinS4t1qbit, line 99: spaces around +

ValentinS4t1qbit (changes requested):
A few implementation suggestions: try to move away from naive python implementations when you can and use numpy when dealing with data-structures of numbers.

A question about design: FCISolver VS get_fci_solver

tangelo/algorithms/classical/fci_solver.py
JamesB-1qbit: We went back and forth many times for Simulator. My understanding was that we changed to `get_backend` mostly for PEP8 reasons. I would prefer to keep both but not use deprecate.

tangelo/algorithms/classical/fci_solver.py
JamesB-1qbit: I did search google for the minimum swaps to order a list. The bottleneck would not be determining the swaps, it would be possibly swapping columns of the psi4.wavefunction.Ca.  

The implementation previously used was not correct as I tested it earlier. I have now added a test for interior frozen orbitals.

I ended up using this one https://www.geeksforgeeks.org/minimum-number-swaps-required-sort-array/. For [1, 0, 2, 5, 6, 7, 3, 4] (which corresponds to freezing orbitals [1, 3, 4]) it returns [(0, 1), (3, 6), (6, 4), (4, 7), (7, 5)]. Your code returns [(0, 1), (2, 2), (3, 5), (3, 6), (4, 6), (4, 7), (5, 7)] which after removing (2,2) requires an extra swap.

AlexandreF-1qbit (approved):

tangelo/algorithms/classical/fci_solver.py
```diff
@@ -131,3 +145,147 @@ def get_rdm(self):
...
+            warnings.warn("Degenerate orbitals are present in the molecule. The fci calculation is performed "
+                          "without symmetry so the rdms may not correspond to the integrals in molecule. A c1 "
+                          "version of the molecule with the correct integrals is present as FCISolverPsi4.molecule.")
+        one_rdm = np.asarray(self.ciwfn.get_opdm(0, 0, "SUM", False))
+        two_rdm = np.asarray(self.ciwfn.get_tpdm("SUM", True))
+
+        return one_rdm, two_rdm
+
+
+fci_solver_dict = {"pyscf": FCISolverPySCF, 'psi4': FCISolverPsi4}
```
ValentinS4t1qbit, line 246: this seems like a very random place for this variable.

tangelo/algorithms/classical/fci_solver.py
```diff
@@ -131,3 +145,147 @@ def get_rdm(self):
...
+
+
+def get_fci_solver(molecule: SecondQuantizedMolecule, solver: Union[None, str, Type[ElectronicStructureSolver]] = default_fci_solver, **kwargs):
+    """Return requested target FCISolver object.
+
+    Args:
+        molecule (SecondQuantizedMolecule) : Molecule
+        solver (string or Type[ElectronicStructureSolver] or None): Supported string identifiers can be found in
+            fci_solver_dict (from tangelo.algorithms.classical.fci_solver). Can also provide a user-defined backend (child to ElectronicStructureSolver class)
+        kwargs: Other arguments that could be passed to a target. Examples are solver type etc.
```
ValentinS4t1qbit, line 256: Still having trouble here: any concrete example ?

Instead of kwargs, I suggest you name them solver_kwargs. What is a "target" here is not obvious for readers, and the body of your function show that current they are just passed to your solver, so it seems more accurate.

tangelo/algorithms/classical/fci_solver.py
```diff
@@ -131,3 +145,147 @@ def get_rdm(self):
...
+            solver = fci_solver_dict[solver.lower()]
+        except KeyError:
+            raise ValueError(f"Error: backend {solver} not supported. Available built-in options: {list(fci_solver_dict.keys())}")
+    elif not issubclass(solver, ElectronicStructureSolver):
+        raise TypeError(f"Target must be a str or a subclass of ElectronicStructureSolver but received class {type(solver).__name__}")
+
+    return solver(molecule, **kwargs)
+
+
+def FCISolver(molecule: SecondQuantizedMolecule, solver: Union[None, str, Type[ElectronicStructureSolver]] = default_fci_solver, **kwargs):
```
ValentinS4t1qbit, line 282: 2 comments:

- I think your code blob is literally equivalent (but less efficient) that writing `FCISolver = get_fci_solver`, since they have the exact same behaviour. You are defining an alias, except you duplicated some code and you have an indirection in the call.

- What's the point, here ? Apart from PEP8 naming convention, why isn't it perfectly suitable for people to use `FCISolver`, which is how all our tests and solvers work ?

We moved from `Simulator` to `get_backend` because we had to move from a legacy name (Backend was more appropriate) and we kept the former around a while to ensure user code would not immediately break (we should totally remove it in the next PR btw).
https://github.com/goodchemistryco/Tangelo/blob/main/tangelo/linq/simulator.py

Here, FCISolver seems to be still perfectly relevant and desirable to me, I don't really see why we have `get_fci_solver`.

JamesB-1qbit: - Not according to my development environment (VSCode). It labels FCISolver as a variable when using `FCISolver = get_fci_solver`, although when you mouse over it, it does return the docstring of a function
Screenshot 2023-06-15 at 10 11 01 AM

-  With the separate function, it labels it as a function, which is correct. There is nothing wrong with it outside of it violating PEP8 so it looks a little strange.
Screenshot 2023-06-15 at 10 11 37 AM

In general, I prefer `get_fci_solver` as this is similar to `get_backend`. We could have just as easily changed `Simulator` to `Backend` but chose `get_backend`.

If we decide that `FCISolver` is the way to go then I will remove `get_fci_solver`.

tangelo/algorithms/classical/fci_solver.py
JamesB-1qbit: It throws an error if I put it at the top of the file. Apparently python cares within a file.

ValentinS4t1qbit: Then I stand corrected :)

Minor but you are not being consistent with `'` and `"` in your dictionary keys.

tangelo/algorithms/classical/fci_solver.py
JamesB-1qbit: Right now I kept both. But now `FCISolver` is a class that has a solver attribute which is `solver` that takes on either `FCISolverPySCF` or `FCISolverPsi4` or user defined. `get_fci_solver` returns `FCISolverPySCF` or `FCISolverPsi4`. We can change the names but I like having both. 

If users want to access attributes without using `fcisolver.solver.whatever`. They can create a solver with `get_fci_solver`.

Where possible, I default to using `IntegralSolverName` but sometimes this is not possible. Especially with the upcoming CCSD solver which ignores frozen orbitals if you try to feed it a reference wavefunction is psi4.

ValentinS4t1qbit (approved):
Thank you James, It took us a few back and forth but I think we got somewhere good !

PR #310, opened 2023-06-13
Support for UMP2 initial parameters

AlexandreF-1qbit:
Support for initial MP2 parameters with a UHF mean-field.

Highlights:
- Support UMP2 initial parameters via `pyscf.mp.UMP2`.
- Moved the computation of MP2 parameters to the `MP2Solver` class (was in `UCCSD` before).
- Fixed a circular import when doing `from tangelo.toolboxes.ansatz_generator._unitary_cc_openshell import uccsd_openshell_get_packed_amplitudes` by moving the `MP2Solver` import after the initialiazation of the `UCCSD` ansatz.

### DIFF
--- a/tangelo/algorithms/classical/mp2_solver.py
+++ b/tangelo/algorithms/classical/mp2_solver.py
@@ -52,7 +52,7 @@ class MP2Solver(ElectronicStructureSolver):
 
         if self.spin != 0 or self.uhf:
             self.n_alpha, self.n_beta = molecule.n_active_ab_electrons
-            self.n_active_moa, self.n_active_mob = molecule.n_active_mos if self.uhf else (molecule.n_active_mos, molecule.n_active_mos)
+            self.n_active_moa, self.n_active_mob = molecule.n_active_mos if self.uhf else (molecule.n_active_mos,)*2
         else:
             self.n_occupied = ceil(molecule.n_active_electrons / 2)
             self.n_virtual = molecule.n_active_mos - self.n_occupied
--- a/tangelo/algorithms/classical/mp2_solver.py
+++ b/tangelo/algorithms/classical/mp2_solver.py
@@ -92,9 +92,9 @@ class MP2Solver(ElectronicStructureSolver):
 
         # Check if MP2 has been performed
         if self.mp2_fragment is None:
-            raise RuntimeError(f"{self.__class__.name}: Cannot retrieve RDM. Please run the 'simulate' method first")
+            raise RuntimeError(f"{self.__class__.__name__}: Cannot retrieve RDM. Please run the 'simulate' method first")
         if self.frozen is not None:
-            raise RuntimeError(f"{self.__class__.name}: RDM calculation is not implemented with frozen orbitals.")
+            raise RuntimeError(f"{self.__class__.__name__}: RDM calculation is not implemented with frozen orbitals.")
 
         one_rdm = self.mp2_fragment.make_rdm1()
         two_rdm = self.mp2_fragment.make_rdm2()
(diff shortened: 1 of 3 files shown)


tangelo/algorithms/classical/mp2_solver.py
```diff
@@ -49,14 +50,25 @@ def __init__(self, molecule):
         self.frozen = molecule.frozen_mos
         self.uhf = molecule.uhf
 
+        if self.spin != 0 or self.uhf:
```
ValentinS4t1qbit, line 53: I need a comment here. What is this code block doing ? What is it providing that wasn't available before ?

tangelo/algorithms/classical/mp2_solver.py
```diff
@@ -85,3 +97,46 @@ def get_rdm(self):
         two_rdm = self.mp2_fragment.make_rdm2()
 
         return one_rdm, two_rdm
+
+    def get_mp2_params(self):
+        """Computes the MP2 initial variational parameters. Compute the initial
```
ValentinS4t1qbit, line 102: I think we need to be careful about our definition and the function name. 

Are we "computing initial parameters for a variational algorithm" here (which sounds like an application, not a definition?) or are we computing single and double amplitudes associated to something that has some chemical meaning, output following a particular convention.... that can be used as initial parameters for variational approaches such as <...> ? 

Ideally you'd want to start with something that explains what we are actually computing before saying what they can be used for.

tangelo/algorithms/classical/mp2_solver.py
```diff
@@ -85,3 +97,46 @@ def get_rdm(self):
...
+        elements into the appropriate convention. MP2 only has doubles (T2)
+        amplitudes, thus the single (T1) amplitudes are set to a small non-zero
+        value and added. The ordering is single, double (diagonal), double
+        (non-diagonal).
+
+        Returns:
+            list of float: The initial variational parameters.
+        """
+
+        # Check if MP2 is performed
```
ValentinS4t1qbit, line 113: Has been performed.

tangelo/algorithms/classical/mp2_solver.py
```diff
@@ -85,3 +97,46 @@ def get_rdm(self):
...
+        if self.mp2_fragment is None:
+            raise RuntimeError("MP2Solver: Cannot retrieve MP2 parameters. Please run the 'simulate' method first")
```
ValentinS4t1qbit, line 115: `"f{self.__class__.name} ..."` for the current class name

tangelo/toolboxes/ansatz_generator/uccsd.py
```diff
@@ -255,22 +251,21 @@ def _get_openshell_qubit_operator(self):
...
         Returns:
             list of float: The initial variational parameters.
         """
 
-        if self.molecule.uhf:
-            raise NotImplementedError(f"MP2 initialization is not currently implemented for UHF reference in {self.__class__}")
         if not is_package_installed("pyscf"):
             raise ValueError(f"pyscf is required for MP2 initial parameters in {self.__class__.__name__}.")
 
+        # Import here to solve an AttributeError: partially initialized module
```
ValentinS4t1qbit, line 264: Could you please post the whole traceback, here ? I want to see what files it's going through before identifying the circular import.

It makes sense for your to import MP2Solver here, I'm wondering what's going on the other side.

This feels more like a comment for the reviewer rather than something you want to leave for the community to read. If we do leave something behind, it will have to be much more concise.

ValentinS4t1qbit (changes requested):
A few things

tangelo/toolboxes/ansatz_generator/uccsd.py
ValentinS4t1qbit: I'm having trouble understanding what lines here triggering the error, and the context. I think it would be easier to have a quick call and running the particular tests live with `python -m unittest <module_name>.<testclass>.<testmethod>`

ValentinS4t1qbit: It seems that python does not handle nested init files the way we expected it to. For now we have agreed to merge this changes and move forward, and then investigate separately. The issue seems to be related to our fundamental understanding of python and imports, and circular dependencies, not Tangelo itself.

ValentinS4t1qbit (approved):

PR #311, opened 2023-06-13
Checkfile for IntegralSolverPySCF

AlexandreF-1qbit:
Support for defining a chkfile during the mean-field calculation. When providing a chkfile name to the `IntegralSolverPySCF` class (`IntegralSolverPySCF("your_chkfile_name.chk")`), it will consider using a file on disk to store mo coefficients / density matrix and other import quantities.

### DIFF
--- a/tangelo/toolboxes/molecular_computation/integral_solver_pyscf.py
+++ b/tangelo/toolboxes/molecular_computation/integral_solver_pyscf.py
@@ -46,7 +46,24 @@ def mol_to_pyscf(mol, basis="CRENBL", symmetry=False, ecp=None):
 
 class IntegralSolverPySCF(IntegralSolver):
     """Electronic Structure integration for pyscf"""
+
     def __init__(self, chkfile=None):
+        """Initialize the integral solver class for pyscf. A chkfile path can be
+        provided.
+
+        Regarding the chkfile, three scenarios are possible:
+        - A chkfile path is provided, but the file doesn't exist: it creates
+            a chkfile at the end of the SCF calculation.
+        - A chkfile path is provided and a file already exists: the initial
+            guess is taken from the chkfile and this file is updated at the end
+            of the calculation.
+        - No chkfile path is provided: The SCF initial guess stays the default
+            one (minao). No chkfile is created.
+
+        Args:
+            chkfile (string): Path of the chkfile.
+        """
+
         from pyscf import gto, lib, scf, symm, ao2mo
         self.gto = gto
         self.lib = lib
--- a/tangelo/toolboxes/molecular_computation/integral_solver_pyscf.py
+++ b/tangelo/toolboxes/molecular_computation/integral_solver_pyscf.py
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import os
+
 import numpy as np
 
 from tangelo.toolboxes.molecular_computation.integral_solver import IntegralSolver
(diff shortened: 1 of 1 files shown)


tangelo/toolboxes/molecular_computation/integral_solver_pyscf.py
```diff
@@ -108,14 +111,21 @@ def compute_mean_field(self, sqmol):
         sqmol.mean_field = self.scf.RHF(molecule) if not sqmol.uhf else self.scf.UHF(molecule)
         sqmol.mean_field.verbose = 0
 
+        if self.chkfile is not None:
```
ValentinS4t1qbit, line 114: `is_chkfile_exist` -> `chkfile_found`

Suggestion: define variable with default value out of branch. Update if necessary.

```
chkfile_found = False
if self.chkfile:
   chkfile_found = os.path.exists(self.chkfile)
   sqmol.mean_field.chkfile = self.chkfile
```

tangelo/toolboxes/molecular_computation/integral_solver_pyscf.py
```diff
@@ -108,14 +111,21 @@ def compute_mean_field(self, sqmol):
...
         # Force broken symmetry for uhf calculation when spin is 0 as shown in
         # https://github.com/sunqm/pyscf/blob/master/examples/scf/32-break_spin_symm.py
-        if sqmol.uhf and sqmol.spin == 0:
+        if sqmol.uhf and sqmol.spin == 0 and not is_chkfile_exist:
             dm_alpha, dm_beta = sqmol.mean_field.get_init_guess()
             dm_beta[:1, :] = 0
             dm = (dm_alpha, dm_beta)
             sqmol.mean_field.kernel(dm)
         else:
+            sqmol.mean_field.init_guess = "chkfile" if is_chkfile_exist else "minao"
```
ValentinS4t1qbit, line 128: So, if a user provides a path to a check file that is incorrect or does not exist, you're not going to tell them or raise an error, instead you're gonna do computations and make them believe everything went according to their plan ?

I'd be tempted to raise a warning or an error as soon as I realize the check file cannot be found. Do you think computing from scratch if it cannot be found is desirable behaviour for users who explicitly tried to provide a chkfile ? In what cases ?

ValentinS4t1qbit (changes requested):
I'm not sure about how we should handle this if the chkfile is not found.

tangelo/toolboxes/molecular_computation/integral_solver_pyscf.py
AlexandreF-1qbit: If there is no chkfile already existing, the code will create one (and the initial guess stays the default one). Without that, the users wouldn't be able to create a chkfile, preventing them to use the feature of getting initial guesses from this file for future calculations. It is actually the expected behaviour.

ValentinS4t1qbit (changes requested):
We need a bit of docs with your explanations from the PR conversation tab (which you can resolve afterwards). Then I believe we are good to merge !

tangelo/toolboxes/molecular_computation/integral_solver_pyscf.py
```diff
@@ -44,13 +46,14 @@ def mol_to_pyscf(mol, basis="CRENBL", symmetry=False, ecp=None):
 
 class IntegralSolverPySCF(IntegralSolver):
     """Electronic Structure integration for pyscf"""
-    def __init__(self):
+    def __init__(self, chkfile=None):
```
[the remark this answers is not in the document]
AlexandreF-1qbit: A docstring should be there now.

ValentinS4t1qbit (approved):
Thank you ! Please think about whether this feature should be advertised or showcased in a way or another. If not particularly, that's fine

PR #313, opened 2023-06-15
CCSD solver psi4

JamesB-1qbit:
Support for CCSD energies only with Psi4. RDMs are not supported.


tangelo/algorithms/classical/tests/test_ccsd_solver.py
```diff
@@ -41,13 +42,22 @@ def test_ccsd_h4_uhf_a1_frozen(self):
...
+        self.assertAlmostEqual(energy, -2.0409800, places=5)
+
     def test_ccsd_be(self):
         """Test CCSDSolver against result from reference implementation."""
 
         solver = CCSDSolver(mol_Be_321g)
         energy = solver.simulate()
 
-        self.assertAlmostEqual(energy, -14.531416589890926, places=6)
+        self.assertAlmostEqual(energy, -14.531416589890926, places=4)
```
ValentinS4t1qbit, line 60: a bit odd seeing 10+ digits (e.g up to questionable machine-precision) while we only care for 5-6.

Can you briefly explain why we relaxed those tests a bit, here?
Is it simply related to the differences between how psi4 and pyscf go about the computation, which are both deemed "correct" ?

ValentinS4t1qbit (changes requested):
A few little things

tangelo/algorithms/classical/tests/test_ccsd_solver.py
JamesB-1qbit: They are both correct to well within chemical accuracy. I assume it may be from the SCF/CCSD convergence threshold or numerical precision of the integrals.

ValentinS4t1qbit (approved):
Great job, thank you ! Another PR down !

PR #317, opened 2023-06-19
add spin to available options and docstring

JamesB-1qbit:
Previous change removed spin from adapt as an option. Added back and added missing argument in docstring


AlexandreF-1qbit (approved):

tangelo/algorithms/variational/adapt_vqe_solver.py
```diff
@@ -93,6 +94,7 @@ def __init__(self, opt_dict):
         self.deflation_circuits: Optional[List[Circuit]] = copt_dict.pop("deflation_circuits", list())
         self.deflation_coeff: float = copt_dict.pop("deflation_coeff", 1)
         self.up_then_down: bool = copt_dict.pop("up_then_down", False)
+        self.spin: Union[int, None] = copt_dict.pop("spin", None)
```
ValentinS4t1qbit, line 97: Should `None` be there ? Docs says `int`. Is there a better default value than `None` ?
Could it be `0` instead ?

tangelo/algorithms/variational/adapt_vqe_solver.py
```diff
@@ -93,6 +94,7 @@ def __init__(self, opt_dict):
         self.deflation_circuits: Optional[List[Circuit]] = copt_dict.pop("deflation_circuits", list())
         self.deflation_coeff: float = copt_dict.pop("deflation_coeff", 1)
```
ValentinS4t1qbit, line 95: could we use `1.` as the default value, to make this clearly a float in the code ?

tangelo/algorithms/variational/adapt_vqe_solver.py
JamesB-1qbit: `n_electron` and `n_spinorbitals` are still None to conform with `fermion_to_qubit_mapping` arguments and checks.

ValentinS4t1qbit: That is ok and whatever we decide to do about this in the future is up to us.
Ideally they should be integers but this is more of a "codebase fix" than something actually needed atm.

ValentinS4t1qbit (approved):
Thank you, this is enough for a quick fix.

I understand that quantities such as spin, n_spinorbitals and n_electrons should be integers, but that we must test whether these have been set by the user or not. `0` is not an ideal default value for these variables, as it is either non-sensical or ambiguous.

I think this fix is ok for now, and that if we really wanted to fix those observations, we'd have to follow where these quantities are going and change a number of statements. It does not sound like this is particularly necessary at this point in time.

PR #319, opened 2023-06-22
skip performance tests when using pytest

JamesB-1qbit:
This test takes a long time and does not print out the times anyway during continuous integration.

Running from terminal `python test_translator_perf.py` still runs the tests.


ValentinS4t1qbit (approved):
Thanks James. Good catch.

PR #320, opened 2023-06-22
Change installation

JamesB-1qbit:
Remove pyscf dependency.

### DIFF
--- a/.github/workflows/run_psi4_test.yml
+++ b/.github/workflows/run_psi4_test.yml
@@ -26,14 +26,12 @@ jobs:
     - name: Install psi4
       shell: bash -el {0}
       run: |
-        conda create -n p4env psi4 -c psi4
+        conda install psi4 python=3.8 -c conda-forge/label/libint_dev -c conda-forge
         conda init
-        conda activate p4env
       if: always()
 
     - name: Install pip, wheel, pytest, jupyter
       run: |
-        conda activate p4env
         python -m pip install --upgrade pip
         pip install wheel
         pip install pytest
@@ -41,27 +39,23 @@ jobs:
     
     - name: Install qulacs
       run: |
-        conda activate p4env
         pip install qulacs
       if: always()
 
     - name: tangelo install
       run: |
-        conda activate p4env
         python -m pip install .
       if: always()
 
     - name: tangelo psi4 integral tests
       run: |
         cd tangelo/toolboxes/molecular_computation/tests
-        conda activate p4env
         pytest --doctest-modules --junitxml=junit/psi4-test-results.xml test_psi4.py
       if: always()
 
     - name: tangelo psi4 classical tests
       run: |
         cd tangelo/algorithms/classical/tests
-        conda activate p4env
         pytest --doctest-modules --junitxml=junit/psi4-classical-test-results.xml
       if: always()
 
--- a/.github/workflows/run_psi4_test.yml
+++ b/.github/workflows/run_psi4_test.yml
@@ -2,13 +2,10 @@ name: psi4 test
 
 on: [pull_request]
 
-env:
-  NO_PYSCF: 1
-
 jobs:
   build:
 
-    runs-on: ubuntu-latest
+    runs-on: windows-latest
     strategy:
       matrix:
         python-version: [3.8]
--- a/.github/workflows/run_psi4_test.yml
+++ b/.github/workflows/run_psi4_test.yml
@@ -5,7 +5,7 @@ on: [pull_request]
 jobs:
   build:
 
-    runs-on: windows-latest
+    runs-on: ubuntu-latest
     strategy:
       matrix:
         python-version: [3.8]
@@ -26,7 +26,7 @@ jobs:
     - name: Install psi4
       shell: bash -el {0}
       run: |
-        conda install psi4 python=3.8 -c conda-forge/label/libint_dev -c conda-forge
+        conda install psi4 python=3.8 -c psi4
         conda init
       if: always()
 
--- a/README.rst
+++ b/README.rst
@@ -96,46 +96,31 @@ separately with ``pip``\ , before trying again.
 With Docker
 ^^^^^^^^^^^
 
-Use our Docker file to deploy Tangelo in a Linux environment, either retrieved from pip or mounted locally.
-Comment / uncomment the relevant sections of the Dockerfile to control installation and dependencies.
+Use our Docker file to deploy Tangelo in a Linux environment, either retrieved from pip or mounted locally. Comment / uncomment the relevant sections of the Dockerfile to control installation and dependencies.
 
 "No install" notebook method
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-A good alternative for users that simply want to quickly get a working environment ready, especially for quick tests, demos, tutorials.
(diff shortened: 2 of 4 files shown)


ValentinS4t1qbit:
I see the main branch test has failed: it's possible that you fixed the target branch afterwards, but Github will only notice and rerun it properly after you push something new. Let's merge the other PRs first, this way you'll update your branches gradually, push, and will naturally fix that.

In the meantime, I'll look into the rest

README.rst
```diff
@@ -124,16 +124,20 @@ Depending on your OS and environment, some of these packages may be more challen
...
 
-Although PySCF is installed by default, this behaviour can be turned off by defining the environment variable `NO_PYSCF=1` and then
-installing using pip by `pip install git+https://github.com/goodchemistryco/Tangelo.git@main` or using the "From source, using setuptools"
-method. 
+The two quantum chemistry packages that are natively supported are `PySCF <https://pyscf.org/>`_ and `Psi4 <https://psicode.org/>`_.
 
-The classical solvers `CCSDSolver`, `MP2Solver`, `FCISolver` and `MINDO3Solver` as well as `DMETProblemDecomposition` require 
-PySCF at this time.
+`DMETProblemDecomposition` is only supported with `PySCF`.
+`SemiEmpircalSolver` requires PySCF  along with 'python3 -m pip install git+https://github.com/pyscf/semiempirical'.
```
ValentinS4t1qbit, line 135: I think we can delete that too. It's easy to figure out, I believe we handle it with ModuleNotFound, and it's only relevant for people who actually call the Semi-Empirical solver and... to my knowledge that's no one so far. Let's not get our already-crowded README more cluttered with minor things.

README.rst
```diff
@@ -124,16 +124,20 @@ Depending on your OS and environment, some of these packages may be more challen
...
 
-Psi4 is supported everywhere except `CCSDSolver`, `MP2Solver`, `FCISolver` and `MINDO3Solver` and `DMETProblemDecomposition`.
+You are also welcome to provide your own interface to a quantum chemistry package of your choice by defining a subclass of
```
ValentinS4t1qbit, line 137: We have a test that illustrates that: instead of trying to explain it, a link to the test I think would give much more. Let's be as concise as possible and reuse our resources rather than "duplicate" or "reformulate" what already exists.

A good question is whether or not we should inject / illustrate the feature in a notebook or not (maybe not worth the hassle). We don't have to mention a notebook or get a notebook ready for this PR. Let's just think about whether it's desirable, write it down here, and resolve this comment.

README.rst
```diff
@@ -124,16 +124,20 @@ Depending on your OS and environment, some of these packages may be more challen
 or any issue regarding the above packages, please check their respective documentation.
 
 
-Having trouble with PySCF?
+Quantum Chemistry Packages
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Tangelo can be used without having a quantum chemistry package installed but many algorithms, by default, depend on one being installed.
```
ValentinS4t1qbit, line 130: I think we need to specify "classical" here, and say what are the main things these package does (compute electronic integrals ? Provide reference classical solvers for HF and post-HF techniques ?)

@AlexandreF-1qbit how would you concisely answer this question ?

README.rst
```diff
@@ -124,16 +124,20 @@ Depending on your OS and environment, some of these packages may be more challen
 or any issue regarding the above packages, please check their respective documentation.
 
 
-Having trouble with PySCF?
+Quantum Chemistry Packages
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
ValentinS4t1qbit, line 128: too many ^^

.github/workflows/run_psi4_test.yml
```diff
@@ -2,13 +2,10 @@ name: psi4 test
...
 on: [pull_request]
 
-env:
-  NO_PYSCF: 1
-
 jobs:
   build:
 
-    runs-on: ubuntu-latest
+    runs-on: windows-latest
```
ValentinS4t1qbit, line 8: I like that we are putting windows into the picture, but now we're a bit all over the place.

- why this change ?
- does that mean we have tests running on ubuntu with pypi , but windows with psi4 ? Do we believe that's what our users are mostly using ?

As a result of this change, are we no longer testing a configuration that we should be testing ?

README.rst
```diff
@@ -124,16 +124,20 @@ Depending on your OS and environment, some of these packages may be more challen
 or any issue regarding the above packages, please check their respective documentation.
 
 
-Having trouble with PySCF?
+Quantum Chemistry Packages
```
ValentinS4t1qbit, line 127: We had another section just before that saying "optional dependencies".
Can we make things flow a bit better ?

It feels like we now have a optional dependency section with a first "sub section" about quantum circuit backends, and another one about classical quantum chemistry backends. Maybe those are the "sub titles" we need, we throw 1-2 links and tell people to look up the installation instructions of these things (e.g it's not our problem)

ValentinS4t1qbit (changes requested):
A question regarding windows for test, and I think our README is quite bloated already: we need to be as smart and concise as possible.

Pretty sure I will make another PR myself to go over this README again (I'm picky) but let's just focus on your additions in this PR.

.github/workflows/run_psi4_test.yml
JamesB-1qbit: I was mostly using it to test whether we need to direct people to use Windows Linux Subsystem. I can change it back. But the tests are now passing and it appears people can use Tangelo in Windows (native) with Psi4.

ValentinS4t1qbit: Great information. Let's stick to Ubuntu now and make the matter of extending things to Windows / Mac / whatever the scope of a subsequent PR, to do things in a smart way.

.github/workflows/run_psi4_test.yml
```diff
@@ -29,42 +26,36 @@ jobs:
     - name: Install psi4
       shell: bash -el {0}
       run: |
-        conda create -n p4env psi4 -c psi4
+        conda install psi4 python=3.8 -c conda-forge/label/libint_dev -c conda-forge
```
ValentinS4t1qbit, line 29: also meh, python 3.8 hardcoded. Let's revert back to the general ubuntu image we had, and we'll try to be smart about windows and such later

README.rst
JamesB-1qbit: I think there should be a notebook to some other package but the link, to the test, I added  should be good enough for now.

README.rst
JamesB-1qbit: I changed to Classical Quantum Chemistry Packages. Perhaps there should be a hyphen or something.

.github/workflows/run_psi4_test.yml
JamesB-1qbit: Psi4 only supports python 3.8 for Windows (native). Reverted

ValentinS4t1qbit (approved):
Cool beans ! I think at this point it's mostly about potential improvements or beautification but we have met the requirements of our release.

If we see anything about:

- testing / Docker
- using environment variables to let people choose a q.chem backend if several are installed or provide a path to it or something useful
- upgrading our documentation or examples
- bug

then it's a good time to talk about it quickly and sneak a PR in. If we don't see anything I think that's fine

PR #325, opened 2023-06-29
Qiskit interface: Bug workaround and PauliSumOp->SparsePauliOp

JamesB-1qbit:
One bug from deepcopy not working anymore with `qiskit.result.marginal_counts` changed to inplace=True. 
We never resample result anyway.

`opflow` is deprecated. `PauliSumOp`->`SparsePauliOp`. See https://qiskit.org/documentation/migration_guides/opflow_migration.html


ValentinS4t1qbit (approved):
Thanks for the fix !

PR #330, opened 2023-07-17
Fix combinatorial mapping when spin!=0

AlexandreF-1qbit:
Fix for the combinatorial mapping when `spin!=0`, i.e. when there are a different number of alpha and beta electrons. For convenience (if we want to work with a matrix), it ensures that the combinatorial integers are between 0 and 2^n_qubit - 1. As of now, when `spin!=0`, `n_choose_alpha != n_choose_beta` and the chosen integers suffer discontinuities. 

This has no effect on the number of qubit and the resulting qubit operators (as they are equivalent by remapping qubit indices). It also has no effect when `n_choose_alpha == n_choose_beta`.


JamesB-1qbit (approved):

ValentinS4t1qbit (approved):
Thank you for the fix ! Does this warrant a new test to be added ? If not, we can merge.

PR #331, opened 2023-08-03
Fixes for MIFNOHelper after QEMIST Cloud 0.6.0 release

AlexandreF-1qbit:
Fixes for the MIFNOHelper class after QEMIST Cloud 0.6.0 release. Outputs have changed, and there is the fragment informations inside the results now. Also, `force_negative_epsilon` has been removed, so it won't cause confusion in the future.


tangelo/problem_decomposition/incremental/mifno_helper.py
```diff
@@ -295,7 +252,7 @@ def compute_fermionoperator(self, molecule, frag_id):
 
         return new_molecule._get_fermionic_hamiltonian(mo_coeff)
 
-    def mi_summation(self, user_provided_energies=None, force_negative_epsilon=False):
+    def mi_summation(self, user_provided_energies=None):
```
ValentinS4t1qbit, line 255: is the negative epsilon thing now not done in Tangelo anymore ? Maybe it happens on QEMIST Cloud instead ?

ValentinS4t1qbit (approved):
Nothing to say :)

ValentinS4t1qbit:
Qiskit has moved and now requires qiskit-aer to be installed separately. That explain most of your test failures.
A change in pyscf changed the behaviour for mean field calculations: that may explain the last one.

James's recent PR for QPE may have fixed all that, I think that if you update your branch the tests may pass.

tangelo/problem_decomposition/incremental/mifno_helper.py
AlexandreF-1qbit: It was mainly for debugging purposes (for small fragments, where MI might be variational), but it shouldn't be enabled to get the same energy as in QEMIST Cloud. However, MI is not variational in a general sense, and I think keeping this may be confusing.

ValentinS4t1qbit (approved):

PR #335, opened 2023-09-05
(Draft) Autocas interface as an IntegeralSolver

AlexandreF-1qbit:
[Autocas](https://scine.ethz.ch/download/autocas) implementation as an IntegralSolver. `get_integrals` still has to be coded (one-body integrals are easily retrieved from the results, but two-body integrals are still missing).


AlexandreF-1qbit:
This PR is left on the ice for some time to get other priorities done.

The current state:
- The autocas inputs are now handled by YAML. An interface with python dictionaries is existing, but only the basis set can be change for a given molecule (https://github.com/qcscine/autocas/blob/master/scine_autocas/__main__.py).
- It seems there is a bug for the basis set settings in YAML (it is ignored).
- We have to develop a way to get two-electron integrals.
- UHF and large CAS computation have not been tested yet.

AlexandreF-1qbit:
A minimal example has been merged into the `Tangelo-Examples` repo (https://github.com/goodchemistryco/Tangelo-Examples/pull/32).

PR #336, opened 2023-09-11
QM/MM support with electrostatic embedding

JamesB-1qbit:
An implementation of QM/MM with electrostatic embedding of the QM region using the partial charges from the MM region.


AlexandreF-1qbit (changes requested):
I got several questions/comments:
- Is there a reason why specifically supporting PDBs?
- I think `MDAnalysis` could be helpful here. It has several I/O function for RDKit, OpenMM and ParmED (https://docs.mdanalysis.org/2.0.0/documentation_pages/converters.html).
- We should ensure there is a path to add "full" QMMM (not only electrostatic embedding). At this point, I don't know how one would tackle this situation.
- Where does the reference values for the test come from?
- I will try to think of a general design before discussing this in a meeting. Maybe I can try to suggest a more scalable/modular alternative (there is many ways of doing QMMMs, i.e. many forcefield or fragmentation, and I would like to be easy to add a custom way). As of now, I think this design is OK.

ValentinS4t1qbit:
I would like us to take this to the finish line whenever we can, this is an important feature.

JamesB-1qbit:
> I got several questions/comments:
> 
> * Is there a reason why specifically supporting PDBs?
> * I think `MDAnalysis` could be helpful here. It has several I/O function for RDKit, OpenMM and ParmED (https://docs.mdanalysis.org/2.0.0/documentation_pages/converters.html).
> * We should ensure there is a path to add "full" QMMM (not only electrostatic embedding). At this point, I don't know how one would tackle this situation.
> * Where does the reference values for the test come from?
> * I will try to think of a general design before discussing this in a meeting. Maybe I can try to suggest a more scalable/modular alternative (there is many ways of doing QMMMs, i.e. many forcefield or fragmentation, and I would like to be easy to add a custom way). As of now, I think this design is OK.

I reformulated the code thanks to your observations. I will respond to your overarching comments below.

*This has been changed such that any list of files should work. As long as the file identifier (the part after the .") is able to tell openbabel which file format is used. It still all runs through pdb as there are many many interfaces. But if people do not like it, they can now use a MMCharges format
*This will require some more work. PySCF(https://pyscf.org/user/solvent.html) and Psi4(https://psicode.org/psi4manual/master/pcmsolver.html) do support implicit solvation in straightforward manner. I think realistically, it appears one can just define a new IntegralSolver for each of these and it's done. I think the new structure does provide a path for explicit solvation but it is certainly more involved.

AlexandreF-1qbit (approved):
Thanks for considering my previous suggestions. I agree with the design as of now, we now have a way to embedded a molecule in an explicit solvent!

I only commented on few points that should be fixed before merging.

tangelo/toolboxes/molecular_computation/mm_charges_solver.py
```diff
@@ -0,0 +1,161 @@
...
+            List[float]: The list of charges
+            List[Tuple(float, Tuple(float, float, float))]: The atoms and geometries of the charges"""
+        pass
+
+
+def convert_files_to_pdbs(filenames: Union[str, List[str]]):
+    """Convert file or list of files to pdb files using openbabel.
+
+    Args:
+        filename (Union[str, List[str]]): The filename(s) that describe the MM region
```
ValentinS4t1qbit, line 42: name of argument is "filenames" according to signature, and it's gonna be written "file name(s)" in your sentence afterwards

tangelo/toolboxes/molecular_computation/mm_charges_solver.py
```diff
@@ -0,0 +1,161 @@
...
+
+    Returns:
+        List[str]: The list of pdb equivalent filenames.
```
ValentinS4t1qbit, line 45: file names

tangelo/toolboxes/molecular_computation/mm_charges_solver.py
```diff
@@ -0,0 +1,161 @@
...
+    elif package.lower() == "rdkit":
+        return MMChargesSolverRDKit()
+    else:
+        raise ValueError(f"Requested MM package of {package} is not currently supported in Tangelo")
+
+
+class MMChargesSolverOpenMM(MMChargesSolver):
+    def __init__(self, force_field_params=('amber14-all.xml', 'amber14/tip3pfb.xml')):
+        self.force_field_params = force_field_params
+        pass
```
ValentinS4t1qbit, line 106: pass means "do nothing" and is only useful if python complains that you forgot the body of a function or indentation does not work out.

Because of line 105, it is not needed here.

tangelo/toolboxes/molecular_computation/mm_charges_solver.py
```diff
@@ -0,0 +1,161 @@
...
+
+        pdbs = [PDBFile(file) for file in pdbfiles]
+        modeller = Modeller(pdbs[0].topology, pdbs[0].positions)
+        for pdb in pdbs[1:]:
+            modeller.add(pdb.topology, pdb.positions)
+        forcefield = ForceField(*self.force_field_params)
+        system = forcefield.createSystem(modeller.topology)
+        nonbonded = [f for f in system.getForces() if isinstance(f, NonbondedForce)][0]
+        charges = [nonbonded.getParticleParameters(i)[0]._value for i in range(system.getNumParticles())]
+        # Obtain xyz string
```
ValentinS4t1qbit, line 125: Don't say "xyz string" but something that makes chemical sense. Did you mean "Generate xyz coordinates of atoms" ? (is it clear what those atoms are from context btw?)

tangelo/toolboxes/molecular_computation/mm_charges_solver.py
```diff
@@ -0,0 +1,161 @@
...
+    def get_charges(self, files) -> List[Tuple[float, Tuple[float, float, float]]]:
+        from openmm.app.pdbfile import PDBFile
+        from openmm.app.forcefield import ForceField
+        from openmm import NonbondedForce
+        from openmm.app import Modeller
+        import openbabel
+
+        pdbfiles = convert_files_to_pdbs(files)
+
+        pdbs = [PDBFile(file) for file in pdbfiles]
```
ValentinS4t1qbit, line 117: Add a comment before that blob and tell me concisely what it builds. Seems to be about a force field and charges (you return the latter).

You can probably group L115 and 117 together, they are purely to generate handles to pdb data but don't do any computation related to your force field.

tangelo/toolboxes/molecular_computation/mm_charges_solver.py
```diff
@@ -0,0 +1,161 @@
...
+            _ = conv.ReadFile(mol, file)
+            geometry = conv.WriteString(mol)
+            geom += atom_string_to_list(geometry.split("\n", 2)[2:][0])
+        return charges, geom
+
+
+class MMChargesSolverRDKit(MMChargesSolver):
+    def __init__(self, mmffVariant="MMFF94"):
+        self.mmffVariant = mmffVariant
+        pass
```
ValentinS4t1qbit, line 140: remove

tangelo/toolboxes/molecular_computation/mm_charges_solver.py
```diff
@@ -0,0 +1,161 @@
...
+
+    def get_charges(self, files) -> List[Tuple[float, Tuple[float, float, float]]]:
+        import rdkit
+        from rdkit.Chem import AllChem
```
ValentinS4t1qbit, line 144: why not just combine l143-144 into `import rdkit.Chem as rdkc` and then just using that directly ?

your code seems to use rdkit.Chem everywhere anyway and you're using `AllChem` only once so its not like importing rdkit entirely or importing AllChem from that submodule helps you.

tangelo/toolboxes/molecular_computation/mm_charges_solver.py
```diff
@@ -0,0 +1,161 @@
...
+            mol_to_add = rdkit.Chem.rdmolfiles.MolFromPDBFile(file, removeHs=False)
+            rdmol = rdkit.Chem.rdmolops.CombineMols(rdmol, mol_to_add)
+        rdkit.Chem.SanitizeMol(rdmol)
+
+        # Read charges and geometry
+        mmff_props = AllChem.MMFFGetMoleculeProperties(rdmol, mmffVariant=self.mmffVariant)
+        charges = [mmff_props.GetMMFFPartialCharge(i) for i in range(rdmol.GetNumAtoms())]
+        rdkit.Chem.SanitizeMol(rdmol)
+        new_xyz = rdkit.Chem.rdmolfiles.MolToXYZBlock(rdmol)
+        # Strip first two lines and convert to standard format
```
ValentinS4t1qbit, line 158: skip a line

tangelo/toolboxes/molecular_computation/tests/test_psi4.py
```diff
@@ -94,6 +96,42 @@ def test_iqcc_h4_uhf(self):
 
         self.assertAlmostEqual(iqcc_energy, -1.95831, places=3)
 
+    def test_oniom_energy_ccsd_hf(self):
```
ValentinS4t1qbit, line 99: oniom ?

tangelo/toolboxes/molecular_computation/tests/test_psi4.py
```diff
@@ -13,17 +13,19 @@
...
 from tangelo import SecondQuantizedMolecule
 from tangelo.helpers.utils import installed_chem_backends
 from tangelo.problem_decomposition.oniom.oniom_problem_decomposition import ONIOMProblemDecomposition
+from tangelo.problem_decomposition.qmmm.qmmm_problem_decomposition import QMMMProblemDecomposition
 from tangelo.problem_decomposition.oniom._helpers.helper_classes import Fragment
 from tangelo.toolboxes.molecular_computation.integral_solver_psi4 import IntegralSolverPsi4
 from tangelo.algorithms.variational import SA_OO_Solver, BuiltInAnsatze, ADAPTSolver, iQCC_solver
 from tangelo.molecule_library import xyz_H4, mol_H4_minao, xyz_H2, mol_H4_sto3g_uhf_a1_frozen
 
+pwd_this_test = os.path.dirname(os.path.abspath(__file__))+"/"
```
ValentinS4t1qbit, line 27: same comment as in other test file: build your alalalala variables here, try to refactor to avoid copy-pasting if its easy, etc

tangelo/toolboxes/molecular_computation/tests/test_psi4.py
```diff
@@ -94,6 +96,42 @@ def test_iqcc_h4_uhf(self):
...
+
+        system = Fragment(solver_high="ccsd", options_low=options_both)
+        qmmm_model_cc = QMMMProblemDecomposition({"geometry": geometry, "qmfragment": system, "charges": charges})
+
+        e_qmmm_cc = qmmm_model_cc.simulate()
+        self.assertAlmostEqual(-98.62087, e_qmmm_cc, places=4)
+
+    def test_energy_fci_h2_ala_ala_ala(self):
+        """Test that the reference energy is returned when an H2 QM geometry is placed next to a pdb charges with VQE as the solver
+        for both rdkit and openmm"""
```
ValentinS4t1qbit, line 115: loop

`for mmpackage in ["rdkit", "open"]:` (...)

tangelo/algorithms/classical/ccsd_solver.py
```diff
@@ -161,17 +161,20 @@ def __init__(self, molecule: SecondQuantizedMolecule):
...
         self.molecule = SecondQuantizedMolecule(xyz=molecule.xyz, q=molecule.q, spin=molecule.spin,
                                                 solver=intsolve,
                                                 basis=molecule.basis,
                                                 ecp=molecule.ecp,
                                                 symmetry=False,
                                                 uhf=molecule.uhf,
                                                 frozen_orbitals=molecule.frozen_orbitals)
+
         self.init_mo_coeff = molecule.mo_coeff
+        self.extra_energy = 0. if not hasattr(molecule.solver, "charges") else self.molecule.solver.ext_pot.computeNuclearEnergy(self.molecule.solver.mol)
```
ValentinS4t1qbit, line 177: same: do we not have a more descriptive name for this "extra energy"?

tangelo/algorithms/classical/ccsd_solver.py
```diff
@@ -21,7 +21,7 @@
 from sympy.combinatorics.permutations import Permutation
 
 from tangelo.toolboxes.molecular_computation.molecule import SecondQuantizedMolecule
-from tangelo.toolboxes.molecular_computation import IntegralSolverPsi4, IntegralSolverPySCF
+from tangelo.toolboxes.molecular_computation import IntegralSolverPsi4, IntegralSolverPySCF, IntegralSolverPsi4QMMM
```
ValentinS4t1qbit, line 24: Note: we do not seem to need `IntegralSolverPySCFQMMM` for some reason.

tangelo/algorithms/classical/fci_solver.py
```diff
@@ -210,10 +214,19 @@ def simulate(self):
             for swap_op in swap_ops:
                 wfn.Ca().rotate_columns(0, swap_op[0], swap_op[1], np.deg2rad(90))
 
+        if hasattr(self, "chrgfield"):
```
ValentinS4t1qbit, line 217: Let's leave a comment about the versions here. I don't know how recent is v1.6 but eventually there will be a todo here and the branch for older versions will be pruned.

ValentinS4t1qbit (changes requested):
Good stuff, nothing big, just comments about docs, spacing, code that can be compressed or even removed.

We'll need a QM/MM notebook and try something that goes beyond basic tests for the November release (e.g not the upcoming release but the next one) in order to really be confident we designed something good able to tackle more ambitious use cases.

tangelo/problem_decomposition/qmmm/qmmm_problem_decomposition.py
```diff
@@ -0,0 +1,161 @@
...
+            raise KeyError(f"Keywords :: {copt_dict.keys()}, not available in {self.__class__.__name__}.")
+
+        # Raise error/warnings if input is not as expected
+
+        if not self.geometry or not self.qmfragment:
+            raise ValueError(f"A geometry and qm fragment must be provided when instantiating {self.__class__.__name__}.")
+
+        self.geometry = atom_string_to_list(self.geometry) if isinstance(self.geometry, str) else self.geometry
+
+        self.distribute_atoms()
```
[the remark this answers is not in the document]
JamesB-1qbit: It can not. distribute_atoms() possibly adds more information `self.mmcharges` depending on how `Fragment` is defined.

tangelo/toolboxes/molecular_computation/integral_solver_pyscf.py
```diff
@@ -292,3 +292,110 @@ def compute_uhf_integrals(self, sqmol, mo_coeff):
         Gpqrs = (two_body_integrals_a, two_body_integrals_ab, two_body_integrals_b)
 
         return hpq, Gpqrs
+
+
+class IntegralSolverPySCFQMMM(IntegralSolverPySCF):
+    """Instantiate Electronic Structure integration with PySCF using electrostatic embedding """
```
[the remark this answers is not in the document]
JamesB-1qbit: Possibly. However, PySCF uses QM/MM (https://pyscf.org/user/qmmm.html) for electrostatic embedding and Solvation (https://pyscf.org/user/solvent.html) models for a polarizability continuum models. So I am comfortable keeping it as QMMM meaning electrostatic embedding.

tangelo/toolboxes/molecular_computation/mm_charges_solver.py
JamesB-1qbit: Oddly enough. If I try to do rdkc.AllChem, it says that it does not exist... `AttributeError: module 'rdkit.Chem' has no attribute 'AllChem'`

tangelo/algorithms/classical/ccsd_solver.py
JamesB-1qbit: We only need it for `IntegralSolverPsi4QMMM` because the frozen orbitals need to be defined before creating the molecule to use CCSD in Psi4. The other `IntegralSolverPsi4` and `IntegralSolverPySCF` are just used to check the `isinstance(molecule.solver)` to determine which CCSDSolver to use.

ValentinS4t1qbit (approved):
Great job, thank you for this new feature !

PR #337, opened 2023-09-11
Bug Fix: n_shots parameter was ignored in get_expectation_value

JamesB-1qbit:
n_shots was ignored in get_expectation_value if not paired with mixed state.

### DIFF
--- a/tangelo/linq/tests/test_simulator.py
+++ b/tangelo/linq/tests/test_simulator.py
@@ -371,7 +371,7 @@ class TestSimulateStatevector(unittest.TestCase):
     def test_get_exp_value_h2_with_n_shots(self):
         """ Get expectation value of large circuits and qubit Hamiltonians corresponding to molecules.
             Molecule: H2 sto-3g = [("H", (0., 0., 0.)), ("H", (0., 0., 0.741377))] with increasing shots
-            obtains better accuracy but never the exact result.
+            reduces standard error as expected.
         """
         qubit_operator = load_operator("mol_H2_qubitham.data", data_directory=path_data, plain_text=True)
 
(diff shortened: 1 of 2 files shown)


tangelo/linq/tests/test_simulator.py
```diff
@@ -368,6 +368,37 @@ def test_get_exp_value_from_statevector_h2(self):
...
+
+        for n_shots in [10**2, 10**4, 10**6]:
+            for b in installed_sv_simulator:
+                sim = get_backend(target=b, n_shots=n_shots)
+                tstart = time.time()
+                energy = sim.get_expectation_value(qubit_operator, abs_circ)
+                tstop = time.time()
+                print(f"H2 get exp value with {b:10s} with n_shots={n_shots} returned {energy:.7f} \t Elapsed: {tstop - tstart:.3f} s.")
+
+                try:
```
ValentinS4t1qbit, line 393: Knowing how some tests failed in the past, I'd much rather we simply compute the 3 values, remove the prints and simply test that error is in descending order. With 100x the amount of shots and no noise that should be granted, and never break due to statistical sampling.

In case the assert fails, we print the name of backend, the 3 values of shots and then the three values of energy diffs, to allow for a visual examination upon failure.

tangelo/linq/tests/test_simulator.py
```diff
@@ -368,6 +368,37 @@ def test_get_exp_value_from_statevector_h2(self):
...
+        qubit_operator = load_operator("mol_H2_qubitham.data", data_directory=path_data, plain_text=True)
+
+        with open(f"{path_data}/H2_UCCSD.qasm", "r") as circ_handle:
+            openqasm_circ = circ_handle.read()
+
+        abs_circ = translate_c(openqasm_circ, "tangelo", source="openqasm")
+        expected = -1.1372704
+        test_fail = False
+
+        for n_shots in [10**2, 10**4, 10**6]:
```
ValentinS4t1qbit, line 385: Let's use a variable `shot_counts = [10**2, 10**4, 10**6]`, that will help with the assert statement I suggest below.

tangelo/linq/tests/test_simulator.py
```diff
@@ -368,6 +368,37 @@ def test_get_exp_value_from_statevector_h2(self):
...
+                tstop = time.time()
+                print(f"H2 get exp value with {b:10s} with n_shots={n_shots} returned {energy:.7f} \t Elapsed: {tstop - tstart:.3f} s.")
+
+                try:
+                    self.assertNotAlmostEqual(energy, expected, delta=1.e-6)
+                    self.assertAlmostEqual(energy, expected, delta=10/np.sqrt(n_shots))
+                except AssertionError:
+                    test_fail = True
+                    print(f"{self._testMethodName} : Assertion failed {b} (result = {energy:.7f}, expected = {expected})")
+            if test_fail:
```
ValentinS4t1qbit, line 399: `assert not test_fail` is better.

I assume at this point every single backend who failed returned a print, but we allowed everything to continue to get all the info, and only that assert decides whether or not the overall test in a success.

ValentinS4t1qbit (changes requested):
I suggest changes for the test, coming from experience of statiscial sampling "randomly" breaking things that are too stringent.

tangelo/linq/tests/test_simulator.py
JamesB-1qbit: I think checking for increasing accuracy may have failed with a fairly high probability. 100x the amount of shots would decrease the std error by a factor of 10. But integrating a normal distribution from -0.1std to 0.1 returns about 0.08. So an 8 percent chance that the 100x fewer shots will land in the std deviation region of the 100x more shots distribution which has about a 33 percent chance of landing outside that region.

I changed the test to calculate the standard deviation of 20 runs. I suspect that this should be enough to obtain the correct result within a factor of 0.8 of the correct standard deviation. The tolerated error is small enough that the bounds are excluding if 100 more shots are taken. The np.allclose test is used to make sure it is not returning the same result close to the exact every time. This was the main failure point of my previous test which would fail about 0.06 percent of the time.

ValentinS4t1qbit (approved):
Thank you !

PR #338, opened 2023-09-12
Classical solver uses mo_coeff from SecondQuantizedMolecule

JamesB-1qbit:
Classical solvers used the molecular coefficients from the initial Hartree-Fock calculation and not the molecular coefficients stored in the SecondQuantizedMolecule object. This fixes it for PySCF and Psi4.

### DIFF
--- a/tangelo/algorithms/classical/tests/test_fci_solver.py
+++ b/tangelo/algorithms/classical/tests/test_fci_solver.py
@@ -15,9 +15,11 @@
 import unittest
 
 import numpy as np
+from openfermion import get_sparse_operator
 
 from tangelo import SecondQuantizedMolecule
 from tangelo.algorithms import FCISolver
+from tangelo.toolboxes.qubit_mappings.mapping_transform import fermion_to_qubit_mapping
 from tangelo.molecule_library import mol_H2_321g, mol_Be_321g, mol_H4_cation_sto3g, mol_H4_sto3g, xyz_H4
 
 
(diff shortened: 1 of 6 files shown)


AlexandreF-1qbit:
This is a good catch, thanks for fixing it.

tangelo/algorithms/classical/tests/test_fci_solver.py
```diff
@@ -81,6 +84,30 @@ def test_fci_H4_interior_frozen_orbitals(self):
...
+        mol = SecondQuantizedMolecule(xyz_H4, frozen_orbitals=[1, 3, 4])
+
+        unitary_mat = np.array([[ 0.99799630, -0.05866044,  0.02358562,  0.00246221],
+                                [ 0.05673690,  0.99556613,  0.07416274,  0.01135272],
+                                [-0.02767097, -0.07212301,  0.99605136, -0.04375241],
+                                [-0.00431649, -0.01432819,  0.04272342,  0.99897486]])
+        # Modify orbitals
+        mol.mo_coeff = unitary_mat.T@mol.mo_coeff
+
+        from openfermion import get_sparse_operator
```
ValentinS4t1qbit, line 101: why did you choose to import these things inside the test and not at the beginning of the module ?

ValentinS4t1qbit (approved):

tangelo/algorithms/classical/tests/test_fci_solver.py
ValentinS4t1qbit: I approved, and as soon as this conversation is resolved (whether accompanied with changes or not) we can merge. Cheers

JamesB-1qbit: I originally was not going to keep this test but realized later that it was good to keep as it confirms that the eigenvalue of FCI and from the FermionOperator are equivalent.

ValentinS4t1qbit (approved):
Merci pour le fix,James !

PR #342, opened 2023-09-19
Support for following VQE minimization in VQESolver

AlexandreF-1qbit:
Highlights:
- HiddenPrints is now set with the `VQESolver.verbose` option.
- Saving the energies via the `save_energies` option. Every `energy_estimation` call can now save the energy (to plot after the simulation).

### DIFF
--- a/tangelo/algorithms/variational/vqe_solver.py
+++ b/tangelo/algorithms/variational/vqe_solver.py
@@ -16,6 +16,7 @@
 electronic structure calculations.
 """
 
+from contextlib import nullcontext
 import warnings
 import itertools
 from typing import Optional, Union, List
@@ -91,6 +92,7 @@ class VQESolver:
         ref_state (array or Circuit): The reference configuration to use. Replaces HF state
             QMF, QCC, ILC require ref_state to be an array. UCC1, UCC3, VSQS can not use a
             different ref_state than HF by construction.
+        save_energies (bool): Flag for saving energy estimation values.
     """
 
     def __init__(self, opt_dict):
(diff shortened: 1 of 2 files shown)


tangelo/algorithms/variational/vqe_solver.py
```diff
@@ -114,6 +116,7 @@ def __init__(self, opt_dict):
         self.verbose: bool = copt_dict.pop("verbose", False)
         self.projective_circuit: Circuit = copt_dict.pop("projective_circuit", None)
         self.ref_state: Optional[Union[list, Circuit]] = copt_dict.pop("ref_state", None)
+        self.save_energies: bool = copt_dict.pop("save_energies", False)
```
ValentinS4t1qbit, line 119: I think this design flows well.

Out of curiosity, if we wanted to also monitor other quantities such as S and S^2 for example, how is that achievable ? Is their computation expensive ?

ValentinS4t1qbit (approved):
A question in comment but apart from that I think this is great.

tangelo/algorithms/variational/vqe_solver.py
AlexandreF-1qbit: This would involve a call to https://github.com/goodchemistryco/Tangelo/blob/293f7610071536ebf0f95076629f74331a79b18e/tangelo/algorithms/variational/vqe_solver.py#L316 at each minimization step. I am expecting this would slow down the process significantly.

ValentinS4t1qbit: No, my comment was following a presentation we both saw at IEEE where someone did that for ADAPT. So it would seem that if we were to monitor these quantities, basic VQE is not where it should happen :)

PR #345, opened 2023-09-27
Simplify method for circuits

ValentinS4t1qbit:
- Added `merge_rotations` method and function, that agglomerates compatible rotations acting on the same qubits
- Added `simplify` method and function that applies cycles of (remove small rotation + merge rotations + remove redundant gates) until no change occur anymore or a max cycle.
- `remove_qubit` option added and set to false: by default qubits with no operations on them are not removed from the circuit to avoid code breaking sneakily. Users can use attributes of the circuit object to identify them and remove manually, or use the trim functions to cut them off a circuit and Hamiltonian before computing an expectation value for instance.

I reused our T3 hands-on on VQE to simply grab the optimal VQE UCCSD circuit for H2 321-g (1200 gates, a few var params below 1e-4, therefore considered small rotations) and applied `simplify` out of the box. It took the circuit down to 700 gates the expectation values are very close (we did remove gates that were not exactly the identity). Took 0.15s on google colab instance. Similarly, big impact on H2O (20K gates), within 5s on a simple google colab instance. I think performance is not an issue since we can already handle such circuits (unless we want this to be streamlined in variational algorithms or something).

Interestingly `pytket` 's `FullPeepHole` takes much longer (impractically longer, about 2 orders of magnitude on these particular use cases) and does not perform as well (see screenshots below).

Screen Shot 2023-09-26 at 6 59 25 PM

Screen Shot 2023-09-26 at 7 24 08 PM

Screen Shot 2023-09-27 at 8 47 48 PM
Screen Shot 2023-09-27 at 7 43 14 PM

### DIFF
--- a/tangelo/linq/circuit.py
+++ b/tangelo/linq/circuit.py
@@ -465,7 +465,8 @@ def merge_rotations(circuit: Circuit):
     gate_qubits = {i: list() for i in range(circuit.width)}
     new_gates = []
 
-    rot_gates = {"RX", "RY", "RZ", "CRX", "CRY", "CRZ"}
+    # TODO: could extend to other variational gates, standard or native to some devices (XX, etc)
+    rot_gates = {"RX", "RY", "RZ", "CRX", "CRY", "CRZ", "PHASE", "CPHASE"}
 
     for gi, gate in enumerate(circuit):
         merge_gate = False
@@ -492,7 +493,7 @@ def merge_rotations(circuit: Circuit):
                 new_gates.append(gate)
                 continue
 
-            # Rotation gate found, combining.
+            # Single rotation gate found common to all qubits: combining.
             if gate.name in rot_gates:
                 if (gate.name, gate.target, gate.control) == (g_prev.name, g_prev.target, g_prev.control):
                     merge_gate = True
--- a/tangelo/linq/circuit.py
+++ b/tangelo/linq/circuit.py
@@ -341,28 +341,55 @@ class Circuit:
             return TypeError("Name of circuit object must be a string")
         return {"name": self.name, "type": "QuantumCircuit", "gates": [gate.serialize() for gate in self._gates]}
 
-    def remove_small_rotations(self, param_threshold=0.05):
+    def remove_small_rotations(self, param_threshold=1e-3, remove_qubits=False):
         """Convenience method to remove small rotations from the circuit.
         See separate remove_small_rotations function.
 
         Args:
-            param_threshold (float): Max absolute value to consider a rotation
-                as a small one.
+            param_threshold (float): Optional, max absolute value to consider a rotation
+                as small enough to be discarded
+            remove_qubits (bool): Optional, remove qubit with no operations assigned left
 
         Returns:
             Circuit: The circuit without small rotations.
         """
-        opt_circuit = remove_small_rotations(self, param_threshold)
+        opt_circuit = remove_small_rotations(self, param_threshold=param_threshold, remove_qubits=remove_qubits)
         self.__dict__ = opt_circuit.__dict__
 
-    def remove_redundant_gates(self):
+    def remove_redundant_gates(self, remove_qubits=False):
         """Convenience method to remove redundant gates from the circuit.
         See separate remove_redundant_gates function.
+
+        Args:
+            remove_qubits (bool): Optional, remove qubit with no operations assigned left
 
         Returns:
             Circuit: The circuit without redundant gates.
         """
-        opt_circuit = remove_redundant_gates(self)
+        opt_circuit = remove_redundant_gates(self, remove_qubits=remove_qubits)
+        self.__dict__ = opt_circuit.__dict__
+
+    def merge_rotations(self):
+        """ Convenience method to merge compatible rotations applied successively on identical qubits indices.
+        The operation is done in-place and alters the input circuit.
+        """
+        opt_circuit = merge_rotations(self)
+        self.__dict__ = opt_circuit.__dict__
+
+    def simplify(self, max_cycles=100, param_threshold=1e-3, remove_qubits=False):
+        """ Convenience method to simplify gates in a circuit, by repeating a set of simple passes
+        until no further changes occurs, or a maximum number of cycles has been reached.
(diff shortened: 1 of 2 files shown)


tangelo/linq/circuit.py
```diff
@@ -415,17 +443,80 @@ def remove_small_rotations(circuit, param_threshold=0.05):
...
+    Returns:
+        Circuit: circuit with merged rotations
+    """
+
+    NoneGate = Gate('NONE', 0)
+
+    gate_qubits = {i: list() for i in range(circuit.width)}
+    new_gates = []
+
+    rot_gates = {"RX", "RY", "RZ", "CRX", "CRY", "CRZ"}
```
AlexandreF-1qbit, line 468: We could add PHASE gate as well (also in remove_small_rotations - maybe a function name change would be necessary).

tangelo/linq/circuit.py
```diff
@@ -415,17 +443,80 @@ def remove_small_rotations(circuit, param_threshold=0.05):
...
+        merge_gate = False
+
+        # Identify qubits the current gate acts on.
+        qubits = gate.target if gate.control is None else gate.target + gate.control
+
+        # For all qubits involved, inspect the previous gate.
+        g_prevs = [gate_qubits[qubit_i][-1][1] if gate_qubits[qubit_i] else NoneGate for qubit_i in qubits]
+
+        # No rotations merged: no single previous gate found operating on all qubits involved.
+        if (NoneGate in g_prevs):
```
AlexandreF-1qbit, line 480: Why creating a `NoneGate` object, `None` would not be enough?

tangelo/linq/tests/test_circuits.py
```diff
@@ -277,13 +278,53 @@ def test_simple_optimization_functions(self):
...
+                  Gate('RZ', 2, parameter=theta + eps)]
+        gates += [Gate('CNOT', 2, 1), Gate('RX', 1, parameter=theta - eps), Gate('CNOT', 1, 0)]
+        gates += [Gate('H', 0), Gate('RX', 1, parameter=pi / 2), Gate('RY', 2, parameter=pi / 2)]
+
+        # Check that everything except the initial X gates can be removed
+        # NB: this removes unecessary qubits as well
+        c = Circuit(gates)
+        c.simplify()
+        assert(c == Circuit([Gate('X', i) for i in range(2)]))
+        print(c.width)
```
AlexandreF-1qbit, line 326: print statement to be removed.

tangelo/linq/circuit.py
```diff
@@ -415,17 +443,80 @@ def remove_small_rotations(circuit, param_threshold=0.05):
     rot_gates = {"RX", "RY", "RZ", "CRX", "CRY", "CRZ"}
     gates = [g for g in circuit._gates if not (g.name in rot_gates and abs(g.parameter) % (2*np.pi) < param_threshold)]
 
-    return Circuit(gates)
+    return Circuit(gates) if remove_qubits else Circuit(gates, n_qubits=circuit.width)
```
AlexandreF-1qbit, line 446: It is a matter of preference (and I also don't recall the original implementation), but I would keep this logic inside the `simply` method. This would remove the `remove_qubits` argument in the `remove_small_rotations` and `remove_redundant_gates` functions. This would ensure that those functions only do 1 thing, i.e. remove small rotations or redundant gates (and would leave the qubit indices as is, thus facilitating debugging if needed).

AlexandreF-1qbit (changes requested):

tangelo/linq/circuit.py
ValentinS4t1qbit: It was the first thing I tried and made the code more complicated, as generic python control flow turned out to implicitly use the `==` comparison between objects, which did not work between `None` and a `Gate` object.

Maybe I could have been smarter.

tangelo/linq/circuit.py
ValentinS4t1qbit: What is the default behaviour you would expect from these "low level functions": remove or not remove qubits ? If a user was to use them directly (not through `simplify`), would you then expect them to "adjust" the number of qubits manually afterwards by deleting or manually ensuring they are all still there ?

The behaviour so far for "low level" functions was removing qubits, since we did not enforce `n_qubits` and let them change lists of gates and rebuild circuits without specifying anything. We had a test confirming that.

tangelo/linq/circuit.py
ValentinS4t1qbit: To clarify: are you asking me to try to leverage the identities

$T^2 = S$ and $S^2 = Z$ ?

AlexandreF-1qbit: Yep, it is the more "general" Phase gate, and we list it in the parametrized gates: https://github.com/goodchemistryco/Tangelo/blob/59f5cb9fb4ae59f0d2bd5991c038a9f00b05d997/tangelo/linq/gate.py#L34. I would guess we could also remove/merge those, as it is the same as performing a rotation in a single-qubit Bloch sphere.

tangelo/linq/circuit.py
AlexandreF-1qbit: Ah I see the problem (the `in` keyword forces a comparison on all objects)! Good to know!

tangelo/linq/circuit.py
ValentinS4t1qbit: I have added PHASE and CPHASE and left XX out of scope for now, as Olivia's team knows IONQ very well and I'd rather have their take on it and leave opportunities for them to contribute :)

tangelo/linq/circuit.py
ValentinS4t1qbit: Moving this forward: 
 
 The past behaviour was to trim necessary qubits, and we were not telling the user. This means that they could end up with shorter bit strings than expected without knowing what happened, or get error from simulators because they now need to trim their Hamiltonian.
 
 I'd rather make this explicit for now at high and low level (this is easy to revert). Moreover the output is then ready to work with the trimming functions implemented by Erika a while back. I think overall this is safer for users and more consistent with our workflow.

PR #358, opened 2023-12-05
Implementation of Frozen Natural Orbitals (FNO)

AlexandreF-1qbit:
(Still in draft mode, tests to be added)

Implementation of Frozen Natural Orbitals framework to truncate the virtual space. Work for the restricted and unrestricted cases.

### DIFF
--- a/tangelo/toolboxes/molecular_computation/fno.py
+++ b/tangelo/toolboxes/molecular_computation/fno.py
@@ -1,3 +1,25 @@
+"""Description from QEMIST Cloud
+In the frozen natural orbital (FNO) method, the virtual molecular orbitals are
+transformed and ranked. The one-particle virtual–virtual block of the 2nd order
+Moller-Plesset density matrix is diagonalized to obtain natural orbitals. The
+eigenvalues are used to truncate the virtual space, while the eigenvectors are
+employed to transform the virtual space.
+
+Paper: arXiv:2002.07901
+
+class LocalizationMethod(IntEnum):
+    NONE = 0
+    BOYS = 1
+    PM = 2
+
+class TruncationMethod(IntEnum):
+    NONE = 0
+    OCCUPANCY = 1
+    PERCENTAGE_OF_OCCUPANCY = 2
+    PERCENTAGE = 3
+    COUNT = 4
+"""
+
 import time
 from functools import reduce
 
@@ -8,9 +30,11 @@ import numpy as np
 
 def do_frozen_core_mp2(mean_field, frozen_occupied):
     """Compute mp2 object and compute correlation energy for full virtual space
+
     Args:
         mean_field: pyscf UHF mean_field object
         frozen_occupied:
+
     Returns: mp2 object and mp2 corr. energy for full virtual space
     """
 
@@ -23,7 +47,8 @@ def do_frozen_core_mp2(mean_field, frozen_occupied):
 
 class FNO(lib.StreamObject):
     def __init__(self, molecule, perturb_theory: mp, frac_fno):
-        """This is FNO class which needs HF, MP2 object along with fraction of FNO occupance %
+        """This is FNO class which needs HF, MP2 object along with fraction of
+        FNO occupance %.
 
         Args:
             molecule (SecondQuantizedMolecule): Molecule object
@@ -36,7 +61,8 @@ class FNO(lib.StreamObject):
         # compute the Fock matrix in new mo basis
         self.fock_ao = molecule.mean_field.get_fock()
         self._mp = perturb_theory
-        # obtain MP2 aplitudes.  Can be unfolded into t2aa, t2ab, t2bb <= t2
+
+        # obtain MP2 aplitudes. Can be unfolded into t2aa, t2ab, t2bb <= t2
         self.t2 = perturb_theory.t2
         # list of occupied orbitals that are frozen in MP2 calculation
         self.frozen_list = perturb_theory.frozen
@@ -63,7 +89,8 @@ class FNO(lib.StreamObject):
             self.tvrt.append(tvrt)
 
     def _compute_mp2_vv_density(self, t2=None):
-        '''Create the virtual-virtual block MP2 density for alpha and beta.'''
+        """Create the virtual-virtual block MP2 density for alpha and beta."""
+
         if t2 is None:
             t2 = self.t2
         t2aa, t2ab, t2bb = t2
@@ -75,11 +102,16 @@ class FNO(lib.StreamObject):
 
         return (dvva + dvva.conj().T, dvvb + dvvb.conj().T)
 
+    def _virtual_virtual_MP2_density(self, mean_field):
+        pass
+
     def _diagonalize_density(self, D, ispin):
         """Diagonalize density and return eval, evec and put no of
         FNO orbitals in appropriate space
+
         Args:
             D (array): density matrix
+
         Returns
             array : eigenvectors of density matrix in appropriate space
         """
@@ -115,6 +147,7 @@ class FNO(lib.StreamObject):
 
     def compute_aa(self, mo_coeff, spin):
         """Compute FNO mo_coeff and mo_energy for given spin"""
+
         fock = self.fock_ao[spin]
         # obtain Fock matrix in mo form
         fock_mo = mo_coeff.T.dot(fock).dot(mo_coeff)
--- a/tangelo/toolboxes/molecular_computation/fno.py
+++ b/tangelo/toolboxes/molecular_computation/fno.py
@@ -1,0 +1,353 @@
+import time
+from functools import reduce
+
+from pyscf import lib, ao2mo, cc, mp, scf
+from pyscf.lib import logger
+import numpy as np
+
+
+def do_frozen_core_mp2(mean_field, frozen_occupied):
+    """Compute mp2 object and compute correlation energy for full virtual space
+    Args:
+        mean_field: pyscf UHF mean_field object
+        frozen_occupied:
+    Returns: mp2 object and mp2 corr. energy for full virtual space
+    """
+
+    pt = mp.UMP2(mean_field).set(frozen=frozen_occupied)
+    pt.verbose = 0
+    fMP2_energy, _ = pt.kernel()
+
+    return pt, fMP2_energy
+
+
+class FNO(lib.StreamObject):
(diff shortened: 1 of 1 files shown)


tangelo/toolboxes/molecular_computation/fno.py
```diff
@@ -0,0 +1,415 @@
...
+class FNO():
+    """Class to interface with the Frozen Natural Orbitals protocol, that aims
+    at reducing the computational cost of a molecular problem by truncating
+    the virtual space. In general, the virtual orbitals are ranked according to
+    their MP2 occupancies, and selected with a given threshold. They are also
+    transformed to the FNO basis, using the eigenvectors of the MP2
+    virtual-virtual density matrix.
+
+    Attributes:
+            sec_mol (SecondQuantizedMolecule): Self-explanatory.
```
ValentinS4t1qbit, line 39: could we `sq_mol` or `sqmol` in this file to be more consistent with the rest of our codebase please?

tangelo/toolboxes/molecular_computation/fno.py
```diff
@@ -0,0 +1,415 @@
...
+        """
+
+        frozen_orbitals = self.get_frozen_indices()
+        sec_mol_updated = self.sec_mol.freeze_mos(frozen_orbitals, inplace=False)
+        sec_mol_updated.mo_coeff = self.mo_coeff
+        return sec_mol_updated
+
+    def compute_fno(self, threshold: Union[float, List[float]]) -> None:
+        """Method to compute and truncate the FNO orbitals. It calls
+        the `_compute_rfno` or the `_compute_ufno`method, whether is appropriate.
```
ValentinS4t1qbit, line 113: whichever

tangelo/toolboxes/molecular_computation/fno.py
```diff
@@ -0,0 +1,415 @@
...
+        Returns:
+            int: Number of FNOs determined by the cumulative sum of fractional
+                occupancies satisfying the provided threshold.
+        """
+
+        fno_occ_cumul_frac_occ = np.cumsum(fno_occ) / np.sum(fno_occ)
+        where_threshold = np.where(fno_occ_cumul_frac_occ <= threshold_frac_occ, 1, 0)
+
+        # +1 to be equivalent to the old function it is based on.
+        number_of_fnos = np.sum(where_threshold ) + 1
```
ValentinS4t1qbit, line 413: remove space before )

tangelo/toolboxes/molecular_computation/fno.py
```diff
@@ -0,0 +1,415 @@
...
+    def compute_fno(self, threshold: Union[float, List[float]]) -> None:
+        """Method to compute and truncate the FNO orbitals. It calls
+        the `_compute_rfno` or the `_compute_ufno`method, whether is appropriate.
+        """
+        self._compute_ufno(threshold) if self.uhf else self._compute_rfno(threshold)
+
+    def get_frozen_indices(self) -> Union[List[int], List[List[int]]]:
+        """Method to determine the indices of the frozen orbitals, and it calls
+        the `_get_restricted_frozen_indices` or the `_get_unrestricted_frozen_indices`
+        method, whether is appropriate.
```
ValentinS4t1qbit, line 120: whichever

tangelo/toolboxes/molecular_computation/fno.py
```diff
@@ -0,0 +1,415 @@
...
+        Args:
+            sec_mol (SecondQuantizedMolecule): The SecondQuantizedMolecule
+                object.
+            threshold (float or list): Threshold(s) for FNO occupancy.
+        """
+
+        # Check if there are frozen virtual orbitals. Print warning that this
+        # setting will be overwritten.
+        if sec_mol.frozen_virtual and sec_mol.frozen_virtual != [[], []]:
+            warnings.warn(f"The frozen orbitals {sec_mol.frozen_virtual} indices are likely to be overwritten in {self.__class__.__name__}.", RuntimeWarning)
```
ValentinS4t1qbit, line 74: "likely" ? If you mean that for sure we are writing over these values (whether or not the new values are the same of different) I think we should just say we overwrite them.

tangelo/toolboxes/molecular_computation/fno.py
```diff
@@ -0,0 +1,422 @@
...
+        # Molecular properties useful for the FNO class methods.
+        self.uhf = sqmol.uhf
+        self.n_mos = self.sqmol.n_mos
+        self.fock_ao = self.sqmol.mean_field.get_fock()
+        self.frozen_occupied = self.sqmol.frozen_occupied
+
+        # Verify threshold format.
+        if isinstance(threshold, float) and 0. < threshold <= 1.:
+            self.threshold = [threshold] * 2 if self.uhf else threshold
+        elif isinstance(threshold, (list, tuple, np.ndarray)) and self.uhf and all(map(lambda x: 0. < x <= 1., threshold)):
```
ValentinS4t1qbit, line 88: It's usually better to force the input to be a specific type, unless you believe that all these types may arise naturally as part of users workflow.

Too many options makes it confusing to the user and can lead to insidious bugs or make maintenance of code / compatibility difficult.

I let you exert your best judgment here and decide whether this is good as it is or not (for example you may know that indeed a tuple could get in there)

tangelo/toolboxes/molecular_computation/fno.py
```diff
@@ -0,0 +1,422 @@
...
+        frozen_indices = np.where(moidx == 0)[0].tolist()
+
+        return frozen_indices
+
+    def _get_unrestricted_frozen_indices(self) -> List[List[int]]:
+        """Method to determine the indices of the frozen orbitals in an
+        unrestricted calculation.
+
+        Returns:
+            List[List[int]]: List containing the frozen orbital indices for
```
ValentinS4t1qbit, line 162: List[List[int]] to me means a list of unknown length containing other lists of unknown lengths. But it sounds like we should maybe return a 2-tuple of List[Int] instead because its very clear.

Making that change shouldn't be difficult since both items support indexing and can be unpacked the same way in the rest of your code.

tangelo/toolboxes/molecular_computation/tests/test_fno.py
```diff
@@ -0,0 +1,101 @@
...
+        """Test the FNO active space selection and the molecular coefficients
+        for the H2/3-21G molecular system, with a restricted mean-field.
+        """
+
+        fno = FNO(mol_H2_321g, 0.50)
+        frozen_orbitals = fno.get_frozen_indices()
+
+        self.assertEqual([2, 3], frozen_orbitals)
+
+        fno_mo_coeff_ref = np.array([
```
ValentinS4t1qbit, line 36: Any chance that this may be up to a sign / permutation in molecular calculations in the mean field or something ? Technically the molecule objects are computed when  the module molecular_library is imported

ignore this comment otherwise.

ValentinS4t1qbit (changes requested):
Left a few comments. Unsure if changes are actually needed, but if not ping me and ill approve.

AlexandreF-1qbit:
Summary of changes:
- `threshold` is now a float. Alpha and beta threshold are expected to be the same. It could be extended if a special use-case needs different threshold per spin.
- I removed typehints. It also removes some dependencies, so I think it is the better alternative (at least before we start deprecating python<3.10, where typehinting is not standard).

tangelo/toolboxes/molecular_computation/fno.py
```diff
@@ -0,0 +1,413 @@
...
+Orbitals (FNOs), to automatically truncate the virtual orbital space.
+
+Reference: arXiv:2002.07901
+"""
+
+import warnings
+
+import numpy as np
+
+from tangelo.algorithms.classical import MP2Solver
```
ValentinS4t1qbit, line 25: note to self: this does not seem to be the case here but I'm concerned that since algorithms depend on molecular computations we are here opening the gates to circular dependencies.

tangelo/toolboxes/molecular_computation/fno.py
```diff
@@ -0,0 +1,413 @@
...
+
+    Attributes:
+            sqmol (SecondQuantizedMolecule): Self-explanatory.
+            uhf (bool): Flag indicating the type of mean field used.
+            n_mos (int): Number of molecular orbitals.
+            fock_ao (np.array): Fock matrix in atomic orbital form.
+            frozen_occupied (list): List of indices of frozen occupied orbitals.
+            threshold (float or list): Threshold(s) for FNO occupancy.
+
+        Properties:
```
ValentinS4t1qbit, line 44: Your indentation here is odd (I would have expected it to be aligned with `Attributes`).

I'm not sure about the doc for properties and methods. 

These will be rendered in the sphinx doc, users can scroll and see them easily (the attributes can be set etc anywhere so its trickier). On top of that you're forcing developers to update your docstring every time a method/property is removed / changed / added.

The description of your methods does not seem very helpful, they are pretty straightforward and its redundant with the method's docstring. Finally, I don't believe this is consistent with the rest of the codebase.

I'm in favour of deleting these two paragraphs, since this is not consistent

tangelo/toolboxes/molecular_computation/fno.py
```diff
@@ -0,0 +1,413 @@
...
+class FNO:
+    """Class to interface with the Frozen Natural Orbitals protocol, that aims
+    at reducing the computational cost of a molecular problem by truncating
+    the virtual space. In general, the virtual orbitals are ranked according to
+    their MP2 occupancies, and selected with a given threshold. They are also
+    transformed to the FNO basis, using the eigenvectors of the MP2
+    virtual-virtual density matrix.
+
+    Attributes:
+            sqmol (SecondQuantizedMolecule): Self-explanatory.
```
ValentinS4t1qbit, line 37: I think you have over-indented the attributes

tangelo/toolboxes/molecular_computation/fno.py
AlexandreF-1qbit: I haven't added FNO to any `__init__.py` files, so I guess it shouldn't be imported when we import something from `molecular_computation`.

ValentinS4t1qbit: I think I still don't fully understand the subtleties of imports, ill have to sit down and watch a video or two

ValentinS4t1qbit (approved):
Thank you Alexandre ! We should add a small bit about FNO in our CChem notebook

PR #360, opened 2023-12-06
UHF reference symmetry labels.

JamesB-1qbit:
Adds the ability to assign symmetry labels for a UHF reference. Needed to change the setter for mo_coef for uhf.

### DIFF
--- a/tangelo/toolboxes/molecular_computation/integral_solver_pyscf.py
+++ b/tangelo/toolboxes/molecular_computation/integral_solver_pyscf.py
@@ -154,10 +154,19 @@ class IntegralSolverPySCF(IntegralSolver):
             raise ValueError("Hartree-Fock calculation did not converge")
 
         if sqmol.symmetry:
-            sqmol.mo_symm_ids = list(self.symm.label_orb_symm(sqmol.mean_field.mol, sqmol.mean_field.mol.irrep_id,
+            if not sqmol.uhf:
+                sqmol.mo_symm_ids = list(self.symm.label_orb_symm(sqmol.mean_field.mol, sqmol.mean_field.mol.irrep_id,
                                                               sqmol.mean_field.mol.symm_orb, sqmol.mean_field.mo_coeff))
-            irrep_map = {i: s for s, i in zip(molecule.irrep_name, molecule.irrep_id)}
-            sqmol.mo_symm_labels = [irrep_map[i] for i in sqmol.mo_symm_ids]
+                irrep_map = {i: s for s, i in zip(molecule.irrep_name, molecule.irrep_id)}
+                sqmol.mo_symm_labels = [irrep_map[i] for i in sqmol.mo_symm_ids]
+            else:
+                sqmol.mo_symm_ids = []
+                sqmol.mo_symm_labels = []
+                for j in range(2):
+                    sqmol.mo_symm_ids.append(list(self.symm.label_orb_symm(sqmol.mean_field.mol, sqmol.mean_field.mol.irrep_id,
+                                                                      sqmol.mean_field.mol.symm_orb, sqmol.mean_field.mo_coeff[j])))
+                    irrep_map = {i: s for s, i in zip(molecule.irrep_name, molecule.irrep_id)}
+                    sqmol.mo_symm_labels.append([irrep_map[i] for i in sqmol.mo_symm_ids[j]])
         else:
             sqmol.mo_symm_ids = None
             sqmol.mo_symm_labels = None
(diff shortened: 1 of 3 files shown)


tangelo/toolboxes/molecular_computation/molecule.py
```diff
@@ -313,9 +313,18 @@ def mo_coeff(self):
     @mo_coeff.setter
     def mo_coeff(self, new_mo_coeff):
         # Asserting the new molecular coefficient matrix have the same dimensions.
-        assert self.solver.mo_coeff.shape == (new_mo_coeff := np.array(new_mo_coeff)).shape, \
-            f"The new molecular coefficients matrix has a {new_mo_coeff.shape}"\
-            f" shape: expected shape is {self.solver.mo_coeff.shape}."
+        if self.uhf:
+            for j in range(2):
+                new_mo_coeff = list(new_mo_coeff)
```
ValentinS4t1qbit, line 318: This does not depend on `j` and is not modified in the loop: are you sure this instruction should be inside the loop or is needed at all ?

There's no docstring with the type of `new_mo_coeff` here so I'm not sure what it is and why it needs to be cast as a list. I think we should have some sort of doc.

tangelo/toolboxes/molecular_computation/molecule.py
```diff
@@ -313,9 +313,18 @@ def mo_coeff(self):
...
+                new_mo_coeff[j] = np.array(new_mo_coeff[j])
+                assert self.solver.mo_coeff[j].shape == new_mo_coeff[j].shape, \
+                    f"The new molecular coefficients matrix for index {j} has a {new_mo_coeff[j].shape}"\
```
ValentinS4t1qbit, line 321: `has shape {...} but the expected shape is {...}`

tangelo/toolboxes/molecular_computation/molecule.py
```diff
@@ -313,9 +313,18 @@ def mo_coeff(self):
...
+                    f" shape: expected shape is {self.solver.mo_coeff[j].shape}."
+            self.solver.mo_coeff = new_mo_coeff
+        else:
+            assert self.solver.mo_coeff.shape == (new_mo_coeff := np.array(new_mo_coeff)).shape, \
+                f"The new molecular coefficients matrix has a {new_mo_coeff.shape}"\
+                f" shape: expected shape is {self.solver.mo_coeff.shape}."
```
ValentinS4t1qbit, line 327: see comment above

tangelo/toolboxes/molecular_computation/molecule.py
```diff
@@ -313,9 +313,18 @@ def mo_coeff(self):
...
+        if self.uhf:
+            for j in range(2):
+                new_mo_coeff = list(new_mo_coeff)
+                new_mo_coeff[j] = np.array(new_mo_coeff[j])
+                assert self.solver.mo_coeff[j].shape == new_mo_coeff[j].shape, \
+                    f"The new molecular coefficients matrix for index {j} has a {new_mo_coeff[j].shape}"\
+                    f" shape: expected shape is {self.solver.mo_coeff[j].shape}."
+            self.solver.mo_coeff = new_mo_coeff
+        else:
+            assert self.solver.mo_coeff.shape == (new_mo_coeff := np.array(new_mo_coeff)).shape, \
```
ValentinS4t1qbit, line 325: Just to be sure, you are using the walrus operator here to do a drive-by assignment so that you can reuse it later in the statement, correct? I haven't used it much yet :)

tangelo/toolboxes/molecular_computation/integral_solver_pyscf.py
```diff
@@ -383,10 +392,18 @@ def compute_mean_field(self, sqmol):
...
+            if not sqmol.uhf:
+                sqmol.mo_symm_ids = list(self.symm.label_orb_symm(sqmol.mean_field.mol, sqmol.mean_field.mol.irrep_id,
                                                               sqmol.mean_field.mol.symm_orb, sqmol.mean_field.mo_coeff))
-            irrep_map = {i: s for s, i in zip(molecule.irrep_name, molecule.irrep_id)}
-            sqmol.mo_symm_labels = [irrep_map[i] for i in sqmol.mo_symm_ids]
+                irrep_map = {i: s for s, i in zip(molecule.irrep_name, molecule.irrep_id)}
+                sqmol.mo_symm_labels = [irrep_map[i] for i in sqmol.mo_symm_ids]
+            else:
+                for i in range(2):
+                    print("mol", sqmol.mean_field.mol.symm_orb)
```
ValentinS4t1qbit, line 402: do we still want this print ?

tangelo/toolboxes/molecular_computation/integral_solver_pyscf.py
```diff
@@ -383,10 +392,18 @@ def compute_mean_field(self, sqmol):
...
-            sqmol.mo_symm_ids = list(self.symm.label_orb_symm(sqmol.mean_field.mol, sqmol.mean_field.mol.irrep_id,
+            if not sqmol.uhf:
+                sqmol.mo_symm_ids = list(self.symm.label_orb_symm(sqmol.mean_field.mol, sqmol.mean_field.mol.irrep_id,
                                                               sqmol.mean_field.mol.symm_orb, sqmol.mean_field.mo_coeff))
-            irrep_map = {i: s for s, i in zip(molecule.irrep_name, molecule.irrep_id)}
-            sqmol.mo_symm_labels = [irrep_map[i] for i in sqmol.mo_symm_ids]
+                irrep_map = {i: s for s, i in zip(molecule.irrep_name, molecule.irrep_id)}
+                sqmol.mo_symm_labels = [irrep_map[i] for i in sqmol.mo_symm_ids]
+            else:
+                for i in range(2):
```
ValentinS4t1qbit, line 401: It is not clear at first glance why this loop is needed, and the use of `i` in your list and dict comprehensions below makes it confusing because it should prevail over it.

Are you sure you need a loop at all, considering the statements below do not use this `i` (I think) ?

ValentinS4t1qbit (changes requested):
Small Readability + docs changes.

tangelo/toolboxes/molecular_computation/molecule.py
JamesB-1qbit: It assigns new_mo_coeff to np.array(new_mo_coeff) in one line so the shape property can be called.

ValentinS4t1qbit (changes requested):
A few very small cosmetic things and comments

ValentinS4t1qbit (approved):

PR #366, opened 2023-12-19
Singlet symmetry-reduced UCCSD

JamesB-1qbit:
Utilizes symmetry to reduce the number of parameters for singlet UCCSD. Also generates the variables

param_dict. Maps the parameter indice to the corresponding spatial orbital excitation
param_dict[0] = (2, 0) means that parameter 1 is utilized for the single excitation (2^0 - 0^2) + (3 ^1 - 1^ 3) for example.

operator_dict maps the excitation to the FermionOperator.

params2keep is the list of parameters from the full packed_amplitudes that are kept in the optimization process. This could be modified by the user if some excitations are known to be small.

### DIFF
--- a/tangelo/algorithms/variational/tests/test_vqe_solver.py
+++ b/tangelo/algorithms/variational/tests/test_vqe_solver.py
@@ -252,6 +252,9 @@ class VQESolverTest(unittest.TestCase):
 
         energy = vqe_solver.simulate()
         self.assertAlmostEqual(energy, -1.97783, delta=1e-4)
+        # Test that the number of variational parameters is less than the total possible number
+        self.assertEqual(vqe_solver.ansatz.n_var_params, 10)
+        self.assertEqual(vqe_solver.ansatz.n_full_var_params, 14)
 
     def test_simulate_qmf_h4(self):
         """Run VQE on H4 molecule, with QMF ansatz, JW qubit mapping, initial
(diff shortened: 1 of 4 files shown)


ValentinS4t1qbit:
Do you see this feature as:

- something that always builds UCCSD ensuring that the number of parameters are minimal (e.g identifying excitations that are mapped to the same parameter, therefore leading to a smaller optimization space that preserves those symmetries)
- something enabling users to manually -enforce- that some excitations are ruled by the same parameter
- both / something else ?

AlexandreF-1qbit:
@ValentinS4t1qbit I think the first point summarize it better, but there is more to it. The forbidden excitations have parameter set to 0 (they are ignored) if the orbital symmetry rule is not totally symmetric (A1).

Therefore the excitations on spin orbitals mapping to the same spatial orbitals are linearly dependant + we can remove excitation we know the parameter will be 0 even after optimization.

ValentinS4t1qbit:
@AlexandreF-1qbit Thank you, that sounds pretty rad. It's actually surprising we haven't implemented something like that a long time ago, since it's so fundamental and impactful for computational cost / convergence. My guess is that we were not using UCCSD much and were mostly working with small systems, which was not that time-consuming.

If we manage to reused cached circuits and the results of the qubit mappings, overall we may end up with a VQE-UCC bundle that can be an order of magnitude faster and more flexible.

AlexandreF-1qbit:
I think it is something that was falling into "nice to have" but was not in line with our research roadmap. It is nice to have shorter UCC circuits, but they still scale badly with the system size.

I can try to answer the second one, since I think his account might be locked (our GitHub auth is passing through okta). `FermionOperator` in OpenFermion have indices that are alternating between alpha (even numbers) and beta (odd numbers) spin orbitals. The fermionic term `i^ j - j^ i` would be linked to `(i+1)^ (j+1) - (j+1)^ (i+1)`.

alexfleury-sb:
This work has been transferred to my repo at https://github.com/alexfleury-sb/Tangelo/tree/sym_uccsd.

PR #371, opened 2024-02-23
Fix: Qiskit v1.0.0

alexfleury-sb:
Highlights:
- Fix changes related to Qiskit 1.0.
- Left `qsharp` out of the tests. Work still to be done to support the QDK.
For [qsharp](https://learn.microsoft.com/en-us/azure/quantum/how-to-migrate-to-modern):
"Most of the in-memory simulators from the Classic QDK have been deprecated, leaving the [sparse simulator](https://learn.microsoft.com/en-us/azure/quantum/sparse-simulator) as the default local simulator in VS Code and the Azure Quantum portal."
- There is an error in PySCF >=2.5.0 when computing MP2 initial parameters for UCCSD. This error isn't showing when using PySCF <= 2.4.0.

### DIFF
--- a/tangelo/helpers/utils.py
+++ b/tangelo/helpers/utils.py
@@ -83,7 +83,7 @@ clifford_backends_simulator = {"stim"}
 
 # Dictionary mapping package names to their identifier in this module
 packages = {p: p for p in all_backends}
-#packages["qdk"] = "qsharp"
+# packages["qdk"] = "qsharp"
 
 # Figure out what is installed in user's environment
 installed_backends = {p_id for p_id, p_name in packages.items() if is_package_installed(p_name)}
--- a/tangelo/helpers/utils.py
+++ b/tangelo/helpers/utils.py
@@ -73,8 +73,8 @@ def deprecated(custom_message=""):
 
 
 # List all built-in backends supported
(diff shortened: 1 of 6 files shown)


ValentinS4t1qbit (approved):

tangelo/helpers/utils.py
```diff
@@ -73,8 +73,8 @@ def new_func(*args, **kwargs):
 
 
 # List all built-in backends supported
-all_backends = {"qulacs", "qiskit", "cirq", "braket", "projectq", "qdk", "pennylane", "sympy", "stim"}
-all_backends_simulator = {"qulacs", "qiskit", "cirq", "qdk"}
+all_backends = {"qulacs", "qiskit", "cirq", "braket", "projectq", "pennylane", "sympy", "stim"}
```
ValentinS4t1qbit, line 76: Can we add `# "qdk"}` at the end of the line as an inline comment ?
Same with line 77.

This way we remember where it used to be if we get back to reactivating it.

ValentinS4t1qbit (changes requested):
Great, just a small change.

Ignore failures for python 3.8 and Pennylane: its normal (https://discuss.pennylane.ai/t/issues-importing-pennylane/3900)

tangelo/helpers/utils.py
alexfleury-sb: Sure!

ValentinS4t1qbit (approved):

PR #373, opened 2024-03-13
Removing Python 3.8 support, and adding 3.12

alexfleury-sb:
The Pennylane dependency is not supporting 3.8 anymore.
Qulacs is currently having installation trouble for 3.12, which they should be working on (https://github.com/qulacs/qulacs/issues/645)

### DIFF
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -9,7 +9,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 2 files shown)


ValentinS4t1qbit (approved):
Some failure with dependencies installs (qulacs), which are a known problem for which an issue was created https://github.com/qulacs/qulacs/issues/645

I think it's acceptable to merge, Tangelo behaves as expected as far as we can tell otherwise.

PR #377, opened 2024-05-15
DMET with HF and MP2 solvers

alexfleury-sb:
This is a low-hanging fruit implementation of HF and MP2 solvers with DMET.

Note: This is untested for the unrestricted MP2, and won't work with UHF.

### DIFF
--- a/tangelo/problem_decomposition/dmet/dmet_problem_decomposition.py
+++ b/tangelo/problem_decomposition/dmet/dmet_problem_decomposition.py
@@ -481,7 +481,7 @@ class DMETProblemDecomposition(ProblemDecomposition):
             solver_fragment = self.fragment_solvers[i]
             solver_options = self.solvers_options[i]
             if solver_fragment == "hf":
-                # This code block would output an error with a UHF mean-field.
+                # This code block would output an error with an UHF mean-field.
                 if self.uhf:
                     raise NotImplementedError("UHF mean-field is not supported for the HF solver.")
                 onerdm = mf_fragment.mo_coeff.T @ mf_fragment.make_rdm1() @ mf_fragment.mo_coeff
--- a/tangelo/problem_decomposition/dmet/dmet_problem_decomposition.py
+++ b/tangelo/problem_decomposition/dmet/dmet_problem_decomposition.py
@@ -87,7 +87,6 @@ class DMETProblemDecomposition(ProblemDecomposition):
 
         self.ao2mo = ao2mo
 
-        default_classical_options = dict()
         default_vqe_options = {"qubit_mapping": "jw",
                                "initial_var_params": "ones",
                                "verbose": False}
@@ -182,7 +181,7 @@ class DMETProblemDecomposition(ProblemDecomposition):
         if not self.solvers_options:
             for solver in self.fragment_solvers:
                 if solver.lower() in {"ccsd", "fci", "mp2", "hf"}:
-                    self.solvers_options.append(default_classical_options)
+                    self.solvers_options.append(dict())
                 elif solver.lower() == "vqe":
                     self.solvers_options.append(default_vqe_options)
                 else:
--- a/tangelo/problem_decomposition/dmet/dmet_problem_decomposition.py
+++ b/tangelo/problem_decomposition/dmet/dmet_problem_decomposition.py
@@ -26,7 +26,7 @@ from tangelo import SecondQuantizedMolecule
 from tangelo.problem_decomposition.dmet import _helpers as helpers
 from tangelo.problem_decomposition.problem_decomposition import ProblemDecomposition
 from tangelo.problem_decomposition.electron_localization import iao_localization, meta_lowdin_localization, nao_localization
-from tangelo.algorithms import FCISolver, CCSDSolver, VQESolver
+from tangelo.algorithms import FCISolver, CCSDSolver, VQESolver, MP2Solver
 from tangelo.toolboxes.post_processing.mc_weeny_rdm_purification import mcweeny_purify_2rdm
 from tangelo.toolboxes.molecular_computation.rdms import pad_rdms_with_frozen_orbitals_restricted, \
     pad_rdms_with_frozen_orbitals_unrestricted
@@ -82,10 +82,12 @@ class DMETProblemDecomposition(ProblemDecomposition):
     def __init__(self, opt_dict):
         if not is_package_installed("pyscf"):
             raise ModuleNotFoundError(f"Using {self.__class__.__name__} requires the installation of the pyscf package.")
-        from pyscf import gto, scf
+        from pyscf import gto, scf, ao2mo
         from tangelo.problem_decomposition.dmet.fragment import SecondQuantizedDMETFragment
-        default_ccsd_options = dict()
-        default_fci_options = dict()
+
+        self.ao2mo = ao2mo
+
+        default_classical_options = dict()
         default_vqe_options = {"qubit_mapping": "jw",
                                "initial_var_params": "ones",
                                "verbose": False}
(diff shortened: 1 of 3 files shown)


tangelo/problem_decomposition/tests/dmet/test_osdmet.py
```diff
@@ -67,6 +67,23 @@ def test_lio2_sto6g_uhf(self):
 
         self.assertAlmostEqual(energy, -156.6243118102, places=4)
 
+    def test_notimplemented_uhf_hf_solver(self):
+        """Tests if setting UHF=True raises an error if a HF solver is selected."""
```
ValentinS4t1qbit, line 71: Just to make sure I understand:

- We do not have a `HFSolver` class. Do you think we should have one in the future ? (how much work is it?)
- Does the `NotImplemented` error have anything to do with DMET in the end, or is it just a HF thing ?

tangelo/problem_decomposition/tests/dmet/test_dmet.py
```diff
@@ -55,6 +55,20 @@ def test_incorrect_number_solvers(self):
         with self.assertRaises(RuntimeError):
             DMETProblemDecomposition(opt_dmet)
 
+    def test_not_implemented_solver(self):
+        """Test
```
ValentinS4t1qbit, line 59: Unfinished ?

tangelo/problem_decomposition/tests/dmet/test_dmet.py
```diff
@@ -55,6 +55,20 @@ def test_incorrect_number_solvers(self):
...
+        """
+
+        opt_dmet = {"molecule": mol_H10_321g,
+                    "fragment_atoms": [1]*10,
+                    "fragment_solvers": "banana",
```
ValentinS4t1qbit, line 64: Just to make sure: this translates as "use the `banana` solver for all 10 fragments" right ?

tangelo/problem_decomposition/tests/dmet/test_dmet.py
```diff
@@ -148,6 +162,37 @@ def test_solver_mix(self):
         energy = solver.simulate()
         self.assertAlmostEqual(energy, -0.94199, places=4)
 
+    def test_solver_hf(self):
+        """Test the energy output of DMET with HF solvers (proof of concept)."""
```
ValentinS4t1qbit, line 166: "A proof of concept (POC) is a demonstration of a product in which work is focused on determining whether an idea can be turned into a reality"

Is doing DMET + HF something new, that no one had attempted before ? Or is it just a regular test ?

tangelo/problem_decomposition/tests/dmet/test_dmet.py
```diff
@@ -148,6 +162,37 @@ def test_solver_mix(self):
...
+                    }
+
+        solver = DMETProblemDecomposition(opt_dmet)
+        solver.build()
+        energy = solver.simulate()
+        self.assertAlmostEqual(energy, mol_H10_321g.mf_energy, places=4)
+
+    def test_solver_mp2(self):
+        """Test the energy output of DMET with MP2 solvers.
+        """
```
ValentinS4t1qbit, line 182: That docstring can be a one-liner

tangelo/problem_decomposition/dmet/dmet_problem_decomposition.py
```diff
@@ -476,17 +478,28 @@ def _oneshot_loop(self, chemical_potential, save_results=False, resample=False,
             # TODO: Changing this into something more simple is preferable. There
             # would be an enum class with every solver in it. After this, we would
             # define every solver in a list and call them recursively.
-            # FCISolver and CCSDSolver must be taken care of, but this is a PR itself.
             solver_fragment = self.fragment_solvers[i]
             solver_options = self.solvers_options[i]
-            if solver_fragment == "fci":
+            if solver_fragment == "hf":
+                # This code block would output an error with an UHF mean-field.
```
ValentinS4t1qbit, line 484: See comment on test file

tangelo/problem_decomposition/dmet/dmet_problem_decomposition.py
```diff
@@ -179,12 +181,12 @@ def __init__(self, opt_dict):
         # If it is a list, we verified that the length is the same.
         if not self.solvers_options:
             for solver in self.fragment_solvers:
-                if solver == "ccsd":
-                    self.solvers_options.append(default_ccsd_options)
-                elif solver == "fci":
-                    self.solvers_options.append(default_fci_options)
-                elif solver == "vqe":
+                if solver.lower() in {"ccsd", "fci", "mp2", "hf"}:
+                    self.solvers_options.append(default_classical_options)
```
ValentinS4t1qbit, line 185: Note: this currently does not do anything, as `default_classical_options` is initialized as an empty dict and never used. L182 says "if there are no solver options passed by the user", L185 says "then we are adding nothing to that". 

- Is that the behaviour you want ?
- Are there classical options common to all these solvers ?

Just want to think a little bit about that, in case we want to put a #TODO in the code, add a comment somewhere or modify this right now.

ValentinS4t1qbit (changes requested):
A few questions, little changes to the code overall

tangelo/problem_decomposition/tests/dmet/test_osdmet.py
alexfleury-sb: - Maybe, for now the `IntegralSolver` classes are kind of occupying this spot. This would require a bit of work, and mainly on the software design side of things. It is not trivial to move things and link them again (i.e. `SecondQuantizedMolecule` + `IntegralSolver.compute_mean_field()`).
- It is a mixed of DMET coding that does not support this as of now + untested science that I am not sure if this is valid. I don't have a use-case as of now to confirm if a solution, when the code will work, is correct.

tangelo/problem_decomposition/tests/dmet/test_dmet.py
alexfleury-sb: Yes

tangelo/problem_decomposition/tests/dmet/test_dmet.py
alexfleury-sb: It is just a test, because we are not considering electron correlation in the fragment, the energy should be equal to the mean-field energy. Considering post-HF methods for some fragments would add correlation energy.

tangelo/problem_decomposition/dmet/dmet_problem_decomposition.py
alexfleury-sb: As of now, we are not supporting solver_options though the instantiation of the Solver classes. This was to keep the `self.solvers_options` attribute the same length vs `self.fragment_solvers`.

tangelo/problem_decomposition/tests/dmet/test_osdmet.py
ValentinS4t1qbit: Thank you. Since "this is complicated", we'll just leave it as that.

tangelo/problem_decomposition/dmet/dmet_problem_decomposition.py
```diff
@@ -182,7 +181,7 @@ def __init__(self, opt_dict):
         if not self.solvers_options:
             for solver in self.fragment_solvers:
                 if solver.lower() in {"ccsd", "fci", "mp2", "hf"}:
-                    self.solvers_options.append(default_classical_options)
+                    self.solvers_options.append(dict())
```
ValentinS4t1qbit, line 184: Could we leave a comment here, stating the purpose you explained earlier (e.g having the same length for something) ?

If you have a suggestion for a # TODO, I think it would be great to leave as comment.

ValentinS4t1qbit (approved):

Issue #387, opened 2024-05-28
Support for multi-controlled rotation (RX, RY, RZ) gates for qiskit backend


ValentinS4t1qbit:
# Issue: Feature Request

We recently went to a hackathon, and participants wanted to solve a challenge using a circuit with multi-controlled Ry gates. They wanted to translate the Tangelo circuit to Qiskit to simulate a fake IBM device or transpile to IBM native gate set, but got a `NotImplemented` error from Tangelo. 

Indeed, we had written that function before Qiskit implemented multi-controlled gates, but Qiskit later supported them:  here's the `mcry` gate https://docs.quantum.ibm.com/api/qiskit/0.43/qiskit.circuit.QuantumCircuit#mcry.

Can you modify https://github.com/ValentinS4t1qbit/Tangelo/blob/main/tangelo/linq/translator/translate_qiskit.py so that multi-controlled gates (CRX, CRY, CRZ) are supported in the translation from Tangelo to Qiskit and remove the error message ? (e.g, a list on integers can be passed for control qubits).

It may be that the `crx`, `cry` and `crz` gates in Qiskit can simply be seen as `mcrx`, `mcry` and `mcrz` with only one control qubit, which may help us resolve things elegantly with very little changes.

**Test**

Making this snippet work for CRY is a good starting point:
```
from tangelo.linq import Gate, Circuit, translate_circuit, get_backend

# Defines a multi-controlled RY gate in tangelo
c = Circuit([Gate('CRY', 2, control=[0,1], parameter=1.)])
c_qiskit = translate_circuit(c, target="qiskit")  # should not return an error

# Qiskit seems to parse this as a sequence of gate
print(c_qiskit)
```

You should see that Qiskit parses that gate into this:
<img width="312" alt="Screenshot 2024-05-25 at 11 52 58 PM" src="https://github.com/goodchemistryco/Tangelo/assets/41597680/ed887cd4-7553-4558-935a-d92b028ffdc9">

To solve this issue:

- Create a test that looks at 3 separate circuits of one multi-controlled gate each (one for CRX, CRY, CRZ, like the one above. We could use a loop to do this elegantly)
- Define the circuit in Tangelo, translate it to qiskit. 
- Write the equivalent native qiskit circuit using their `mcr*` gate.
- Test that the translated qiskit object is "equal" to the native one.

golanor:
Hi, I've submitted a PR to address this, but I'm unsure if I should base on main or develop. What should I do?

[...]

PR #390, opened 2024-05-29
Added support for multi-controlled gates on qiskit

golanor:
Added support for multi-controlled gates on qiskit, see #387


tangelo/linq/tests/test_translator_circuit.py
```diff
@@ -180,6 +180,26 @@ def test_qiskit(self):
         sim_results = qiskit_simulator.run(translated_circuit).result()
         np.testing.assert_array_almost_equal(sim_results.get_statevector(translated_circuit), reference_big_msq, decimal=6)
 
+    @unittest.skipIf("qiskit" not in installed_backends, "Test Skipped: Backend not available \n")
+    def test_qiskit_multi_control(self):
+        from tangelo.linq import Gate, Circuit, translate_circuit
```
ValentinS4t1qbit, line 185: All of these have been imported already at the beginning of the module, I think they can be removed (L25)

ValentinS4t1qbit (changes requested):
Thank you, this looks good ! You even supported an extra gate ! The test is convincing. 

- I think some imports can be cleaned up.
- One of the checks currently fails (`pycodestyle` says theres a trailing white space somewhere). Some other checks fail but I don't think it is related to your contribution and we know why.

Out of curiosity: would considering the CRX, CRY, etc as MCRX, MCRY, ... with a single control qubit be possible ? This would simplify the code, and then we may not even need the `crx`, `cry`, `crz`, gate from Qiskit in the dictionary.

It may be that if a user passes a single integer in the Tangelo gate, the software automatically converts it into a list of 1 element. In that case, the the full list of control qubits can always be passed in the translation.

If I am wrong and this becomes tricky, we can forget about this improvement.

golanor:
> Out of curiosity: would considering the CRX, CRY, etc as MCRX, MCRY, ... with a single control qubit be possible ? This would simplify the code, and then we may not even need the `crx`, `cry`, `crz`, gate from Qiskit in the dictionary.
> 
> It may be that if a user passes a single integer in the Tangelo gate, the software automatically converts it into a list of 1 element. In that case, the the full list of control qubits can always be passed in the translation.
> 
> If I am wrong and this becomes tricky, we can forget about this improvement.

It depends on the end realization - the compilation of the MCR() gates is wasteful compared to the CR() gates. If that doesn't matter then we can probably change it.

Regarding the trailing whitespace - I haven't found where it is happening, or which action fails. Can you please point me to where to find it?

ValentinS4t1qbit:
> It depends on the end realization - the compilation of the MCR() gates is wasteful compared to the CR() gates. If that doesn't matter then we can probably change it.

Then I do not mind leaving them alone, the code is still pretty simple. Thanks for the insight.

> Regarding the trailing whitespace - I haven't found where it is happening, or which action fails. Can you please point me to where to find it?

Unexpected: in an unrelated file outside of your PR, I see a whitespace in the printing format. Our linter complains about it.
Would you please remove that whitespace ? This way all tests should pass and we're calling this PR done !

../tangelo/algorithms/variational/adapt_vqe_solver.py:322:61: E203 whitespace before ':'

I saw that while checking the list of checks at the bottom of this PR page, and clicking "details".
Out of curiosity, are you able to click and access the details of the checks ? It would be troublesome if our users could not see them.

Screenshot 2024-05-29 at 6 26 56 PM

Thank you so much, appreciate your patience on this. Unexpected things happen :)

golanor:
> Unexpected: in an unrelated file outside of your PR, I see a whitespace in the printing format. Our linter complains about it. Would you please remove that whitespace ? This way all tests should pass and we're calling this PR done !
> 
> ../tangelo/algorithms/variational/adapt_vqe_solver.py:322:61: E203 whitespace before ':'
> 
> I saw that while checking the list of checks at the bottom of this PR page, and clicking "details". Out of curiosity, are you able to click and access the details of the checks ? It would be troublesome if our users could not see them.

> Thank you so much, appreciate your patience on this. Unexpected things happen :)

I only now realized I can scroll down the checks and see the ones that failed... 

Thank you for being so available! I fixed the whitespace issue.

tangelo/linq/tests/test_translator_circuit.py
```diff
@@ -180,6 +180,26 @@ def test_qiskit(self):
         sim_results = qiskit_simulator.run(translated_circuit).result()
         np.testing.assert_array_almost_equal(sim_results.get_statevector(translated_circuit), reference_big_msq, decimal=6)
 
+    @unittest.skipIf("qiskit" not in installed_backends, "Test Skipped: Backend not available \n")
+    def test_qiskit_multi_control(self):
+        from tangelo.linq import translate_circuit
```
ValentinS4t1qbit, line 185: This function was already imported as `translate_c` earlier in the file, we can just reuse that and remove this line :)

ValentinS4t1qbit (changes requested):
A small final change and the bounty is yours ! 

Your import was already available as a slightly different name, we do not need imports inside that function (in general we prefer to import at the beginning of the module, rather than re-importing inside each function that needs it).

The checks have all passed, so I'm sure it will all work out after your next change.

ValentinS4t1qbit (approved):
Looks good to me !

Thank you for your contribution, I hope this was an interesting experience.
The devil's in the details, and details accumulate over time and make open-source codebase challenging... thank you for your patience :)

If you think Tangelo is an interesting project, please don't hesitate to give us a little ⭐ a consider mentioning it or using it :)

Issue #386, opened 2024-05-28
Augment the Rotosolve optimizer to support Rotoselect


ValentinS4t1qbit:
# Issue: Feature Request

The `Rotosolve` optimizer is a straightforward method to obtain the optimal value for a single qubit rotation in variational quantum algorithms. The current implementation in Tangelo supports its use in [ongoing research](https://arxiv.org/abs/2211.10501) involving the Qubit Coupled Cluster algorithm where typically only a single rotation axis is considered. This feature can be easily augmented to the `Rotoselect` optimizer, which chooses  both the optimal angle and axis of single qubit rotation.

- Paper describing Rotoselect: https://arxiv.org/pdf/1905.09692
- Implementation of Rotosolve in Tangelo: https://github.com/goodchemistryco/Tangelo/blob/main/tangelo/toolboxes/optimizers/rotosolve.py

**Describe the solution you'd like**

The solution would include a separate solver called `Rotoselect` which modifies the rotation axis (RX, RY, RZ) of the gates in the variational circuit. For each rotation axis, it could call `rotosolve_step` to find the optimal angle.

**Finish line**

Testing for this feature should be analogous to the [rotosolve tests](https://github.com/goodchemistryco/Tangelo/blob/main/tangelo/toolboxes/optimizers/tests/test_rotosolve.py).  The test would require an input variational circuit, and a Hamiltonian to minimize. These should be constructed such that the output optimized circuit has a different rotation axis for at least one rotation gate.

cburdine:
Hi Valentin,

I'm here for UnitaryHack and thought this might be a good problem for me to tackle. Could you assign me to this issue? Thanks!

ValentinS4t1qbit:
Game on, @cburdine ! Do your best :)

I recommend you work on something that has the desired behaviour and tries to reuse some `Rotosolve` code, and then ping @elloyd-1qbit and I to have a quick look at it and give you feedback. Then you should be able to have a clear checklist to take you to the finish line and earn the bounty :)

ValentinS4t1qbit:
Solved by #392 by @cburdine

Issue #384, opened 2024-05-27
Circuit as reference state in the ansatz definition


alexfleury-sb:
# Circuit as reference state in the ansatz definition

**What problem does this feature request help you overcome?**

In our stack of Ansätze, located in `toolboxes/ansatz_generator`, most of them take a string `reference_state` as an initialization parameter. This determines the reference state of the ansatze, which is commonly `HF` for an Hartree-Fock state or `zero` for an empty circuit. As an example, here is the code that handles the `reference_state` for the `toolboxes.ansatz_generator.uccsd.UCCSD` ansatz.

https://github.com/alexfleury/Tangelo/blob/9079cb2e8c4b893027a7ff70101e968548000714/tangelo/toolboxes/ansatz_generator/uccsd.py?plain=1#L136-L153

This is restrictive in situations where non-conventional mapping are used, i.e. where `HF` has no meaning.

**Describe the solution you'd like**

A workaround would be to add support for a [tangelo.linq.circuit.Circuit](https://github.com/alexfleury/Tangelo/blob/9079cb2e8c4b893027a7ff70101e968548000714/tangelo/linq/circuit.py#L32) objects as a reference state. Logics have to be implemented to handle the right number of qubits in some edge cases, and to consider optimization (or not) if the reference circuit contains variational gates.

For some ansatze, it would remove the need of specifying the `mapping` and `up_then_down` parameters, as they are only used for reference state preparation. 

As a minimal use-case, exploring combinatorial mapping optimization could be easier if this option is available. For e.g., if the code below could work without error, it would be awesome!

```python
from tangelo.molecule_library import mol_H2_sto3g
from tangelo.toolboxes.qubit_mappings import combinatorial
from tangelo.toolboxes.operators import count_qubits
from tangelo.linq import Circuit, Gate
from tangelo.toolboxes.ansatz_generator import HEA

H = combinatorial(mol_H2_sto3g.fermionic_hamiltonian, mol_H2_sto3g.n_active_mos, mol_H2_sto3g.n_active_electrons)
n_qubits = count_qubits(H)

ref_circ = Circuit([Gate("RY", 0, parameter=0.5), Gate("RY", 1, parameter=-0.5)])
ansatz = HEA(n_qubits=n_qubits, n_layers=2, reference_state=ref_circ)
ansatz.build_circuit()
```

cburdine:
Hi @alexfleury-sb and @ValentinS4t1qbit ,

[...]

PR #392, opened 2024-05-31
Unitaryhack issue 386: Augment the Rotosolve optimizer to support Rotoselect

cburdine:
This is a PR to address [Issue 386: Augment the Rotosolve optimizer to support Rotoselect](https://github.com/goodchemistryco/Tangelo/issues/386)

### :sparkles: New Features:
 *  Added a `rotoselect()` function that implements Algorithm 2 from [https://arxiv.org/pdf/1905.09692](https://arxiv.org/pdf/1905.09692). This implementation requires 7 function evaluations (per parameter per iteration).
* Added an `extrapolate_exp()` function to extrapolate sinusoidal expectation values using the methods from Appendix A in the above paper.
* Added unit tests for the `rotoselect()` algorithm (Heisenberg model and a trivial single-qubit Hamiltonian).

### :construction: Changes to Existing Features:
* Modified the `rotosolve_step()` to return both the optimized parameter set and the extrapolated estimate of the expectation value at that optimal parameter set. This modification is necessary for code reuse, and to allow for further optimization of the existing implementation of `rotoselect()`.
* Added an additional `extrapolate` option to the `rotoselect()` algorithm to allow for the use of extrapolated expectation values from the last call to `rotosolve_step()` in place of an actual function evaluation. This optimization setting reduces the number of function evaluations from three (per parameter per iteration) to two (per parameter per iteration). The `extrapolate` argument is currently set to False by default. More details about this optimization below

---

### :memo: Footnote on optimizing `rotosolve()`:

The existing `rotosolve_step()` function used $\phi = 0$ by default, which is reasonable because it allows for the removal of the gate at compile time. However, if $\phi = \theta_{d-1}^\ast$ is used, then the extrapolated expectation value of $\langle M \rangle_{\theta_{d-1}^\ast}$ returned from the previous call to `rotosolve_step()` can be used in place of a direct evaluation of it. `rotosolve_step()` has been modified so that the gate angle $\phi$ and a previously extrapolated expectation value $\langle M \rangle_{\theta_{d-1}^\ast}$ can be passed into `rotosolve_step()` via the optional arguments `phi` and `m_phi` respectively. By selecting $\phi = \theta_{d-1}^\ast$ and using the extrapolated $\langle M \rangle_{\phi} = \langle M \rangle_{\theta_{d-1}^\ast}$ from the previous optimized parameter, we only need to evaluate $\langle M \rangle_{\phi + \pi/2}$ and $\langle M \rangle_{\phi + \pi/2}$ to optimize the current parameter. In total, this requires 2/3 the number of evaluations, as long as the extrapolated value is accurate.

On an ideal quantum computer with no sampling error, this optimized `rotosolve()` (i.e. with `extrapolate=True`) should behave identically to the previous unoptimized version. However, I anticipate that it may perform slightly worse than the unoptimized `rotosolve()` on noisy devices for the same number of iterations. While the unoptimized version (i.e. with `extrapolate=False`) is more stable, it requires 50% more evaluations than the optimized version. In terms of the total number of evaluations (which is probably the best basis of comparison), I am not sure which will perform better under noise. If in the future you find that the optimized version is too unstable for noisy devices, then you may consider removing the `extrapolate` argument.


ValentinS4t1qbit:
@cburdine Thank you so much for the PR and the detailed description ! @elloyd-1qbit and I will take a look at it.

In the meantime, it seems our linter (`pycodetest`) picked up some stuff and will make your tests fail. If you ask for details, and open the menus you'll see the code files and lines that were brought up.

It seems at first glance all unit tests pass though, which is rad :)

tangelo/toolboxes/optimizers/tests/test_rotosolve.py
```diff
@@ -17,7 +17,8 @@
 import numpy as np
 
 from tangelo.linq import get_backend, Gate, Circuit
-from tangelo.toolboxes.optimizers.rotosolve import rotosolve
+from tangelo.toolboxes.optimizers.rotosolve import rotosolve, rotoselect
+from tangelo.toolboxes.operators.operators import QubitHamiltonian
```
ValentinS4t1qbit, line 21: @alexfleury-sb Would you recommend here we use `QubitOperator` instead ? I believe we haven't followed through on `QubitHamiltonian`: we currently have a class that exists but is not really in use ? It seems here that what is used in the assert statement in the text is energy

alexfleury-sb: Yes, I think the behaviour of `QubitHamiltonian` and `QubitOperator` should be the same. We never used it to its full potential, and it can cause hiccups in some situations. This is why staying with `QubitOperator` is safer.

tangelo/toolboxes/optimizers/rotosolve.py
```diff
@@ -15,10 +15,36 @@
 import numpy as np
 
 
-def rotosolve_step(func, var_params, i, *func_args):
+def extrapolate_expval(theta, m_0, m_minus, m_plus, phi=0.0):
+    """Extrapolates the expectation value of an observable
+    M with respect to a single parameterized rotation (i.e.
+    RX, RY, RZ) with angle theta. The extrapolation is
+    computed based on samples taken at the angles
+    phi, phi+pi/2, and phi-pi/2. This function used the
```
alexfleury-sb, line 23: uses

alexfleury-sb (approved):
Great PR! I'll weight in for the previous question about `ansatz.update_var_gates()`. I think this would be a useful function, and I prefer this approach to adding `gates` to `ansatz.update_var_params()`. I don't know if you plan to add restrictions, i.e. only on rotation gates, but I would personally leave a lot of freedom (whatever makes more sense for you, we usually leave the burden of validating the circuit to the user).

elloyd-1qbit:
Well done! I agree adding a `ansatz.update_var_gates()` function would be really handy.

elloyd-1qbit (approved):

tangelo/toolboxes/optimizers/rotosolve.py
cburdine: In commit [88c127f](https://github.com/goodchemistryco/Tangelo/pull/392/commits/88c127fe8a997d1f1d510892dcd40c9f94b4b4af) I have changed the docstring to:

*The extrapolation uses samples taken at the angles phi, phi+pi/2, and phi-pi/2.*

tangelo/toolboxes/optimizers/tests/test_rotosolve.py
cburdine: In commit [3a4b907](https://github.com/goodchemistryco/Tangelo/pull/392/commits/3a4b907d3360008385c159d00df91b835277be29) I have changed the tests to use `QubitOperator`.

ValentinS4t1qbit:
@cburdine To clarify: do we intend to add `ansatz.update_var_gates()` in this PR ? Does it feel straightforward enough to implement ?

Or would the recommendation be opening a "feature suggestion" issue ? I know you're working on the hackathon, I want to be mindful of your time.

cburdine:
It looks like adding an `update_var_gates()` function will require a change to the interface in the abstract class `Ansatz`.  Functions will then need to be implemented for each ansatz class individually.  Test cases will also need to be added to each of the ansatzes, which may take some time. (If you can think of a better way of doing this, let me know! :bulb:)

 I'm currently working on getting the first draft of a PR for [issue 384](https://github.com/goodchemistryco/Tangelo/issues/384) submitted, which is similar in that it requires extending the functionality of each ansatz class individually. Once I get that submitted I will circle back to this PR and add this feature.

ValentinS4t1qbit (approved):
Great work, thank you so much for your contribution !

As you have touched Tangelo quite a bit with those PRs, any feedback would be very welcome, I wonder how you felt about using it. Any issue (bug, feature request) is very welcome as well.

PR #394, opened 2024-05-31
Fixing Pyscf Incompatibility problem 

### LINKED ISSUE #383: Incompatibility with PySCF >= 2.5.0
# Incompatibility with PySCF >= 2.5.0

This error currently occurs in a single test: [ansatz_generator.tests.test_uccsd.UCCSDTest.test_uccsd_H4_open](https://github.com/goodchemistryco/Tangelo/blob/9079cb2e8c4b893027a7ff70101e968548000714/tangelo/toolboxes/ansatz_generator/tests/test_uccsd.py#L107-L127), where unrestricted MP2 amplitudes are computed as initial parameters for a UCCSD ansatz in VQE. 

With PySCF<=2.4.0, the test passes. However, with the latest versions of PySCF (2.5.0 and on) the test currently returns an error: it seems there was a change in a the `eri` data-structures. It may be that a simple fix could be done inside our `MP2Solver` class.

**Expected Behavior**
Find a way to make this test pass for both PySCF 2.5.0 while still supporting the older versions. Test locally in both environment (you can easily switch between versions with pip install and uninstall, by specifying the desired version).
The continuous integration environment yml file should then be bumped to pyscf latest version so that all automated tests run with it on.

**Current Behavior**
An `IndexError: too many indices for array: array is 1-dimensional, but 2 were indexed` is raised when computing MP2 initial parameters for open-shell systems.

**Steps to Reproduce (minimal example)**

Minimal script:
```python
from tangelo.molecule_library import mol_H4_cation_sto3g
from tangelo.toolboxes.ansatz_generator.uccsd import UCCSD

uccsd_ansatz = UCCSD(mol_H4_cation_sto3g)
uccsd_ansatz.build_circuit()
```

outputs
```shell
Traceback (most recent call last):
  File "/Users/alexandre.fleury/Work/Scratch/sandbox.py", line 6, in <module>
    uccsd_ansatz.build_circuit()
  File "/Users/alexandre.fleury/Repos/Tangelo/tangelo/toolboxes/ansatz_generator/uccsd.py", line 164, in build_circuit
    self.set_var_params()
  File "/Users/alexandre.fleury/Repos/Tangelo/tangelo/toolboxes/ansatz_generator/uccsd.py", line 127, in set_var_params
    initial_var_params = self._compute_mp2_params()
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/alexandre.fleury/Repos/Tangelo/tangelo/toolboxes/ansatz_generator/uccsd.py", line 283, in _compute_mp2_params
    mp2.simulate()
  File "/Users/alexandre.fleury/Repos/Tangelo/tangelo/algorithms/classical/mp2_solver.py", line 316, in simulate
    return self.solver.simulate()
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/alexandre.fleury/Repos/Tangelo/tangelo/algorithms/classical/mp2_solver.py", line 89, in simulate
    _, self.mp2_t2 = self.mp2_fragment.kernel()
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/alexandre.fleury/.pyenv/versions/qsdk/lib/python3.11/site-packages/pyscf/mp/mp2.py", line 598, in kernel
    eris = self.ao2mo(mo_coeff)
           ^^^^^^^^^^^^^^^^^^^^
  File "/Users/alexandre.fleury/.pyenv/versions/qsdk/lib/python3.11/site-packages/pyscf/mp/ump2.py", line 437, in ao2mo
    return _make_eris(self, mo_coeff, verbose=self.verbose)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/alexandre.fleury/.pyenv/versions/qsdk/lib/python3.11/site-packages/pyscf/mp/ump2.py", line 499, in _make_eris
    eris._common_init_(mp, mo_coeff)
  File "/Users/alexandre.fleury/.pyenv/versions/qsdk/lib/python3.11/site-packages/pyscf/mp/ump2.py", line 475, in _common_init_
    mo_a = mo_coeff[0][:,mo_idx[0]]
IndexError: too many indices for array: array is 1-dimensional, but 2 were indexed
```

The `IndexError` points out the fact that the `mo_coeff` data structures have changed from PySCF 2.4.0 to 2.5.0. The solution as of now is to restrict the PySCF support to 2.4.0 and older versions, but it would be great if this bug could be fixed for future releases.

anushkrishnav:
Fixes #383 

The problem inherently lies with PYSCF which they have corrected a few weeks ago, for the time being, my fixes will help remove the issues with the tests, this should be future-proof since it checks if there is any changed in mf_coeff so even if later psf version rolls out with the fix , the code would still work  as there must be a change in mf_coeff which will then not trigger the if condition

had a lot of fun working on this issue, initially though it was something to do with the data structure of mo_coeff then saw it was a pyscf issue 

Hopefully this would help close the issue  for now

### DIFF
--- a/tangelo/algorithms/classical/mp2_solver.py
+++ b/tangelo/algorithms/classical/mp2_solver.py
@@ -85,13 +85,12 @@ class MP2SolverPySCF(ElectronicStructureSolver):
             self.mp2_fragment = self.mp.UMP2(self.mean_field, frozen=self.frozen)
         else:
             self.mp2_fragment = self.mp.RMP2(self.mean_field, frozen=self.frozen)
-            if self.mean_field.istype('UHF'):
-                self.mp2_fragment.mean_field = self.mean_field.to_uhf()
-
-        
-
+            if self.mp2_fragment.mo_coeff.shape == self.mean_field.mo_coeff.shape:
+                mf = self.mean_field.to_uhf()
+                self.mp2_fragment = self.mp.UMP2(mf, frozen=self.frozen, mo_coeff=mf.mo_coeff, mo_occ=None)
 
         self.mp2_fragment.verbose = 0
+        print(self.mp2_fragment.mo_coeff)
         _, self.mp2_t2 = self.mp2_fragment.kernel()
 
         total_energy = self.mp2_fragment.e_tot
--- a/tangelo/algorithms/classical/mp2_solver.py
+++ b/tangelo/algorithms/classical/mp2_solver.py
@@ -81,9 +81,15 @@ class MP2SolverPySCF(ElectronicStructureSolver):
 
         # Execute MP2 calculation
         if self.uhf:
+
             self.mp2_fragment = self.mp.UMP2(self.mean_field, frozen=self.frozen)
         else:
             self.mp2_fragment = self.mp.RMP2(self.mean_field, frozen=self.frozen)
+            if self.mean_field.istype('UHF'):
+                self.mp2_fragment.mean_field = self.mean_field.to_uhf()
+
+        
+
 
         self.mp2_fragment.verbose = 0
         _, self.mp2_t2 = self.mp2_fragment.kernel()


ValentinS4t1qbit:
Maybe @alexfleury-sb has suggestions on how to handle the fix, thanks to your investigation. Alexandre, the automated tests currently force PySCF 2.4.0: do we have a clever way to enable @anushkrishnav to run the test suite on 2.5.0 in his PR ?

@anushkrishnav Before giving you a review, I think you should look at the "checks" at the bottom of this page, click "details" and see what tests are failing and why. Right now, the failures suggest that things are not working for whatever version of pyscf is used in the automated tests (see above).

It is btw perfectly acceptable to have a conditional statement that literally tests the version of pyscf at runtime and does 2 different things depending on the version detected (e.g before and after 2.5.0?). I'd rather be clear and explicit about why things are done, otherwise we'll end up with unexpected instructions for which we do not really remember the purpose.

alexfleury-sb:
I am fairly positive that changing https://github.com/goodchemistryco/Tangelo/blob/6791dc83d9dac7ee5aad9d5a1f3dcbed81aac2cb/.github/workflows/continuous_integration.yml#L76 would indeed change the pyscf in the CI for this specific PR. We could change it to something like `pyscf>=2.5.0` to make sure it is not going back to 2.4.0 for whatever reasons I can't think of at the moment (unless there are mechanisms in the code to handle the pyscf versions at runtime).

anushkrishnav:
The investigation took the most time tbh, 

> It is btw perfectly acceptable to have a conditional statement that literally tests the version of pyscf at runtime and does 2 different things depending on the version detected (e.g before and after 2.5.0?). I'd rather be clear and explicit about why things are done, otherwise we'll end up with unexpected instructions for which we do not really remember the purpose. (I can add a comment on that to make things clear )

In the issue I have detailed whats going on So we don't really need a before and after, the whole thing was cause by a bug from pyscf side , which has now been fixed, I guess the next version will have that fix, this issue that we had was not a feature rather a bug from their side

ValentinS4t1qbit:
@anushkrishnav Thank you for investigating this ! Sorry for the delay, weekend + international travels.

Local testing : I think an easy way to test if your code works for both is simply to test in your local environment one version, then uninstall pyscf with `pip uninstall -y pyscf` and then install the other version with `pip install pyscf==x.y.z`. We do not know what versions of pyscf our users rely on and do not know when the new pyscf version with the fix will be released: i'd rather we have a conditional statement for now (`if pyscf.__version__ < 2.5.0` or something like that).

Automated tests on Github: In order for automated tests to run with the desired version of pyscf, you can modify the yml workflow file in `.github/workflows/continuous_integration.yml`. If you scroll down, you'll see that @alexfleury-sb enforced `pip install pyscf==2.4.0` after we saw the latest version was troublesome. Here I suggest you replace it by `pip install pyscf` to get the latest (this way when the fix is released, it will be used).

Your goal is for these tests to pass !

anushkrishnav:
> @anushkrishnav Thank you for investigating this ! Sorry for the delay, weekend + international travels.
> 
> Local testing : I think an easy way to test if your code works for both is simply to test in your local environment one version, then uninstall pyscf with `pip uninstall -y pyscf` and then install the other version with `pip install pyscf==x.y.z`. We do not know what versions of pyscf our users rely on and do not know when the new pyscf version with the fix will be released: i'd rather we have a conditional statement for now (`if pyscf.__version__ < 2.5.0` or something like that).
> 
> Automated tests on Github: In order for automated tests to run with the desired version of pyscf, you can modify the yml workflow file in `.github/workflows/continuous_integration.yml`. If you scroll down, you'll see that @alexfleury-sb enforced `pip install pyscf==2.4.0` after we saw the latest version was troublesome. Here I suggest you replace it by `pip install pyscf` to get the latest (this way when the fix is released, it will be used).
> 
> Your goal is for these tests to pass !

I have already tested mine on local, I installed and uninstalled the package to see how it performs. I will now move to the automated tests, is it something that I can do ? updating the workflow?

ValentinS4t1qbit:
> I have already tested mine on local, I installed and uninstalled the package to see how it performs. I will now move to the automated tests, is it something that I can do ? updating the workflow?

Yes, that is what the 2nd paragraph of my previous answer was suggesting

anushkrishnav:
Updated the CI file

anushkrishnav:
Not sure how do I navigate the issue, there has not been any change except the one made to get it back to older state

tangelo/algorithms/classical/mp2_solver.py
```diff
@@ -83,7 +83,17 @@ def simulate(self):
         if self.uhf:
             self.mp2_fragment = self.mp.UMP2(self.mean_field, frozen=self.frozen)
         else:
-            self.mp2_fragment = self.mp.RMP2(self.mean_field, frozen=self.frozen)
+            import pyscf
+            if pyscf.__version__ == '2.5.0' and self.mean_field.istype('ROHF'):
```
ValentinS4t1qbit, line 87: I was expecting `>= 2.5.0`, I think we can assume that the change is here to last over newer versions as well

tangelo/algorithms/classical/mp2_solver.py
```diff
@@ -83,7 +83,17 @@ def simulate(self):
...
+                mf = self.mean_field
+                mf = mf.remove_soscf()
+                mf = mf.to_uhf()
+                if getattr(mf, 'with_df', None):
+                    self.mp2_fragment = self.mp.UMP2(mf, frozen=self.frozen, mo_coeff=mf.mo_coeff, mo_occ=None)
+                else:
+                    self.mp2_fragment = self.mp.UMP2(mf, frozen=self.frozen, mo_coeff=mf.mo_coeff, mo_occ=None)
```
ValentinS4t1qbit, line 94: This looks exactly like L92 to me, I am a bit confused why we have that conditional statement

ValentinS4t1qbit (changes requested):
Thank you so much for working on this !

- you can definitely ignore the `MINDO3` test currently failing, this has nothing to do with your change, we'll create an issue for it :)
- I don't quite understand some of your changes. @alexfleury-sb would you mind having a look to help him take this to the finish line ?

tangelo/algorithms/classical/mp2_solver.py
anushkrishnav: The thing is, pyscf made a mistake in the code in version 2.5 which has been fixed now but the fix will be available only from the later version so this issue is very specific to the 2.5 version after which we wont need this fix

tangelo/algorithms/classical/mp2_solver.py
anushkrishnav: Yeshh they are very similar I took it from the pyscf code and placed it , wanted to ask about this, I don't think the result is any different for both case

tangelo/algorithms/classical/mp2_solver.py
ValentinS4t1qbit: That's true, I forgot you mentioned that, thank you.

tangelo/algorithms/classical/mp2_solver.py
ValentinS4t1qbit: If these lines are exactly the same, let's just remove L91-93 and keep L94. If everything goes well, by the time we get a version of pyscf with the fix, this will all be trivial :)

tangelo/algorithms/classical/mp2_solver.py
anushkrishnav: Yehh

tangelo/algorithms/classical/mp2_solver.py
anushkrishnav: Will fix this

alexfleury-sb (approved):
@ValentinS4t1qbit This is a good solution to our previous problems. As I read in the thread, the bug was fixed on pyscf side and is only appearing with version 2.5.0. 

For the MINDO tests that fail, I can take care of them! I think this is out of scope for this PR.

anushkrishnav:
> @ValentinS4t1qbit This is a good solution to our previous problems. As I read in the thread, the bug was fixed on pyscf side and is only appearing with version 2.5.0.
> 
> For the MINDO tests that fail, I can take care of them! I think this is out of scope for this PR.

Awesome , So hopefully this is a wrap for the problem ?

anushkrishnav:
Thank you for the help and guidance @ValentinS4t1qbit

ValentinS4t1qbit (approved):
Thank you so much for the hard work @anushkrishnav !

If you have any suggestions or feedback after playing a bit with this 🍊 Tangelo 🍊 , please don't hesitate ! Otherwise, just leaving us a ⭐ on Github if you think this project is interesting to support us :)

PR #416, opened 2025-12-19
Address semgrep security issue in create_release_branch.yml workflow

k0walik:
This pull request introduces two changes:

- It updates `download-artefact` and `upload-artefact` actions to version v4. This change is necessary because Github Actions workflows (`Continuous Integration` and `psi4`) are failing due to the use of a deprecated version.
- It addresses a semgrep issue in `create_release_workflow.yml`. Previously, using variable interpolation `${{…}}` with `github` context data in a `run:` step could have allowed an attacker to inject their own code into the runner. To prevent this, we should use an intermediate environment variable with `env:` to store the data, and then use the environment variable in the `run:` script.

### DIFF
--- a/.github/workflows/continuous_integration.yml
+++ b/.github/workflows/continuous_integration.yml
@@ -89,19 +89,19 @@ jobs:
       if: always()
 
     - name: Upload nopyscf test results
-      uses: actions/upload-artifact@v3
+      uses: actions/upload-artifact@v4
       with:
         name: tangelo-no-pyscf-test-results
         path: tangelo/toolboxes/molecular_computation/tests/junit/nopyscf-test-results_${{ matrix.python-version }}.xml
 
     - name: Upload pytest test results
-      uses: actions/upload-artifact@v3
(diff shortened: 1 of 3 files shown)


k0walik:
@ValentinS4t1qbit, you may want to have a look why psi4 tests are failing. Quick look shows the computed and expected values are not the same.

alexfleury-sb (approved):
Hello @k0walik, thanks for overlooking the security aspect of the repo.

The tests that are failing are related to H4, which we had problems in the past because of symmetry reasons (similar orbitals get ordered differently sometimes, which caused different energy output). 

The fact that the python 3.9 test is passing, and the other ones don't tells me that maybe there were updates in the psi4 repo that changes the stochastic behaviour of the orbital ordering, and this update might not be available for older pythons. Since we haven't maintained this repo actively, I can't tell for sure if this is the problem.

The best course of actions for now would be to ignore these tests, as it requires some work to point the root cause.

ValentinS4t1qbit (approved):
Thank you !

PR #419, opened 2026-05-26
[linq] Circuit convenience methods return self (closes #408)

### LINKED ISSUE #408: [BUG] <insert description>
# Issue: Bug Report

When using `Circuit.simplify` on a Circuit where all gates simply (ex: a circuit with 1 qubit, 2 consecutive H gates), the result is a `NoneType`.

**Expected Behavior**

It should instead be a circuit with 0 gates, containing the same number of qubits as before `simplify`.

**Current Behavior**
Tell us what happens instead of the expected behavior. Try to be clear and concise if you can. Feel free to add screenshots.

**Steps to Reproduce (minimal example)**

`Circuit([Gate("H", 0)]*2).simplify()` returns `NoneType`

**Possible Solution**

If the simplified circuit contains zero gates, `simplify` needs to return a circuit of 0 gates, containing the same number of qubits as the input circuit.

999purple999:
Closes #408.

`Circuit.simplify()` returns `NoneType` instead of an empty circuit, because the convenience wrapper splats the underlying function's result into `self.__dict__` and returns nothing. Three sibling methods on `Circuit` have the exact same shape, so this PR fixes all four at once for consistency.

## What changed

**`tangelo/linq/circuit.py`** — added `return self` to:
- `Circuit.simplify()`
- `Circuit.remove_redundant_gates()`
- `Circuit.remove_small_rotations()`
- `Circuit.merge_rotations()`

The two docstrings that already promised `Returns: Circuit: The circuit without ...` are tightened; the two that didn't get a matching `Returns` clause added.

**`tangelo/linq/tests/test_circuits.py`** — new regression test `test_simplify_returns_circuit_when_everything_cancels` that pins the exact failing one-liner from the issue body:

```python
result = Circuit([Gate("H", 0)] * 2).simplify()
assert result is not None
assert isinstance(result, Circuit)
assert len(result._gates) == 0
assert result.width == 1
```

Plus analogous assertions for the other three methods.

**`FIX_ISSUE_408_README.md`** — design rationale (why `return self` over a separate `simplified()` API), full file listing.

## Compatibility

The in-place semantic is preserved — these methods still mutate `self.__dict__` before returning. The existing `test_simplify_circuit` passes unchanged because the assertions on circuit state after `.simplify()` are still valid.

The change enables the chainable expression form (`Circuit(...).simplify().draw()`) without breaking any existing code that called these methods for their side-effects and discarded the return value.

## Open question for review

If you'd rather have a separate `simplified()` / `remove_redundant_gates_copy()` API (returning a new circuit, leaving the original untouched), this PR can be reworked in that direction — happy to align on whichever shape fits the linq module's broader conventions.

---

Francesco Pernice Botta · 999purple999

### DIFF
--- a/tangelo/linq/circuit.py
+++ b/tangelo/linq/circuit.py
@@ -379,52 +379,78 @@ class Circuit:
         """Convenience method to remove small rotations from the circuit.
         See separate remove_small_rotations function.
 
+        The operation is performed in place; ``self`` is returned so the call
+        also reads naturally as an expression
+        (e.g. ``c2 = Circuit(...).remove_small_rotations()``).
+
         Args:
             param_threshold (float): Optional, max absolute value to consider a rotation
                 as small enough to be discarded
             remove_qubits (bool): Optional, remove qubit with no operations assigned left
 
         Returns:
-            Circuit: The circuit without small rotations.
+            Circuit: ``self`` after the simplification pass, even when the
+                resulting circuit contains zero gates.
         """
         opt_circuit = remove_small_rotations(self, param_threshold=param_threshold, remove_qubits=remove_qubits)
         self.__dict__ = opt_circuit.__dict__
+        return self
 
     def remove_redundant_gates(self, remove_qubits=False):
         """Convenience method to remove redundant gates from the circuit.
         See separate remove_redundant_gates function.
 
+        The operation is performed in place; ``self`` is returned so the call
+        also reads naturally as an expression.
+
         Args:
             remove_qubits (bool): Optional, remove qubit with no operations assigned left
 
         Returns:
-            Circuit: The circuit without redundant gates.
+            Circuit: ``self`` after the simplification pass, even when the
+                resulting circuit contains zero gates.
         """
         opt_circuit = remove_redundant_gates(self, remove_qubits=remove_qubits)
         self.__dict__ = opt_circuit.__dict__
+        return self
 
     def merge_rotations(self):
-        """ Convenience method to merge compatible rotations applied successively on identical qubits indices.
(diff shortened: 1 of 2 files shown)


alexfleury-sb:
Thanks for the PR @999purple999! This is still in draft mode, so I am assuming your work is not completed yet.

Code changes are pretty much good with me. Two things:
- I would like is that this PR targets the `develop` branch instead of the `main` one.
- Please remove the `FIX_ISSUE_408_README.md` file (and maybe post the content directly in issue #408 for archiving purposes).

999purple999:
Thanks @alexfleury-sb! Both points addressed in 1eb1da6:
- Rebased onto develop (PR base also changed via gh).
- Removed FIX_ISSUE_408_README.md from the PR.

The README content (root-cause walk-through + design rationale for `return self`) can be archived as a comment on #408 if helpful  happy to post it there separately, just say the word. Otherwise it lives in the commit message + PR description.

Let me know if anything else needs a tweak.

999purple999:
Hi @alexfleury-sb — both changes applied in commit [1eb1da6](https://github.com/999purple999/Tangelo/commit/1eb1da6):

- Branch rebased onto `develop` (single commit, no merge of unrelated work).
- `FIX_ISSUE_408_README.md` removed from the diff.

The PR is no longer marked as draft. Ready for re-review whenever you have a moment — thanks for taking the time on this one.

alexfleury-sb (approved):
LGTM

ValentinS4t1qbit (approved):
Sorry, we lost sight of this ! thank you :)

PR #421, opened 2026-08-04
ci: remove the dead workflow-based Semgrep scan (App already covers this repo)

paul-abb:
Removes `.github/workflows/semgrep.yml`. It is dead code, and it is the only thing keeping a stale credential relevant.

## Why

**It has not run in over a year.** GitHub auto-disabled it (`state: disabled_inactivity`); the last successful run was July 2025.

**Semgrep already covers this repo.** The Semgrep GitHub App is installed org-wide and posts the `semgrep-cloud-platform/scan` check on pull requests here — including on the currently open PRs in this repo. Deleting this workflow removes no coverage; the App-based scan is the org's preferred path.

**It is the only consumer of a long-lived token.** The workflow's sole step reads `secrets.SEMGREP_APP_TOKEN`, a repository secret last rotated in January 2025. The App-based scan needs no such token. With this file gone the secret has no consumer and can be removed and revoked — tracked internally.

## Also worth noting about the file being deleted

It was never hardened: unpinned `actions/checkout@v4`, an unpinned `semgrep/semgrep` container image, and no `permissions:` block (so it inherited the org default of a writable `GITHUB_TOKEN`). Rather than fix three things in a workflow that is disabled and redundant, this removes it.

## Validation

- Confirmed `disabled_inactivity` and the last-run dates via the Actions API.
- Confirmed the `semgrep-cloud-platform/scan` check is present and passing on open PRs in this repo, so App coverage is live, not assumed.
- Confirmed `SEMGREP_APP_TOKEN` is referenced nowhere else in this repository.
- No other workflow references this file.


paul-abb:
### `check_branch` fails here by design — it is not a defect in this PR

`protect_main.yaml` enforces a repo policy: a PR into `main` must come from `develop`.

```yaml
if: github.base_ref == 'main' && github.head_ref != 'develop'
run: |
  echo "ERROR: You can only merge to main from develop."
  exit 1
```

This PR has to target `main`, because `.github/workflows/semgrep.yml` **only exists on `main`** —
it is absent from `develop` (`GET contents/.github/workflows/semgrep.yml?ref=develop` → 404).
Retargeting to `develop` would make the PR a no-op, so `check_branch` cannot be made green
without violating the policy it enforces.

Two ways to land it, both a maintainer call:

1. Admin-merge this PR, accepting the expected `check_branch` red.
2. Close this and fold the deletion into the next `develop` → `main` reconciliation. `main` is
   currently 8 ahead / 14 behind `develop`, so that reconciliation is due anyway.

The `build (3.x)` reds are separate and also pre-existing — same dependency drift documented in
#420, unrelated to deleting a workflow file.

For context on why the file should go: the workflow's state is `disabled_inactivity` (GitHub
auto-disabled the cron) and its last run was 2025-07-05. Semgrep coverage is unaffected —
scanning here is provided by the Semgrep Cloud integration rather than by this workflow, which
is why `semgrep-cloud-platform/scan` is green on this very PR while the workflow is dead.

